PR #58, opened 2015-01-19
Refactor macros and add warnings for common issues

tindzk:
Splits pickling macros into separate functions. There are
now warnings for the following three cases:

1) Trait is unsealed
2) Trait does not have any sub-classes
3) No companion symbol found for case classes

We are unfortunate to be affected by the two compiler bugs
SI-7046 and SI-7567 which are the cause for 2) and 3). Yet,
the developer can apply simple workarounds as described in
the warnings.

The current suggestion for 2) is to use integer constants:

```scala
type Base = Int  // sealed trait Base
object Base {
  val Child1 = 0 // case object Child1 extends Base
  val Child2 = 1 // case object Child2 extends Base
}
```

A type-safe variant would be to use an approach similar
to https://github.com/lloydmeta/enumeratum/.

As for 3) there may be even a workaround that could be
implemented in uPickle. A possible solution is described
here:

http://stackoverflow.com/questions/16440124/get-the-companion-object-instance-of-a-inner-modul-with-the-scala-reflection-api

### DIFF
--- a/shared/main/scala/upickle/Macros.scala
+++ b/shared/main/scala/upickle/Macros.scala
@@ -1,11 +1,9 @@
 package upickle
 
 import scala.reflect.macros._
-import scala.reflect._
-import scala.annotation.{ClassfileAnnotation, StaticAnnotation}
+import scala.annotation.StaticAnnotation
 import scala.language.experimental.macros
 
-//import acyclic.file
 /**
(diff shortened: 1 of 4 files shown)


lihaoyi:
Merged, with tweaks that provide a better error message on scala 2.11 using whitebox macro. Thanks a lot! This will go out with Scala.js 0.6.0 when that's released

Issue #59, opened 2015-01-19
Omit field keys in serialised ADTs


tindzk:
As mentioned in #55, I'd like to use uPickle to parse an [aeson](https://hackage.haskell.org/package/aeson)-produced file. Another interoperability issue I encountered is that uPickle doesn't encode field keys. Take the following example:

```scala
case class Format(wrapped: String)
case class RawBlock(format: Format, str: String)

RawBlock(Format("html"), "<center>")
```

Here is a comparison of the differences:

*aeson*
```json
{"t":"RawBlock","c":["html","<center>"]}
```

*uPickle*
```json
["RawBlock",{"format":{"wrapped":"html"},"str":"<center>"}]
```

As long as a class constructor doesn't use any optional parameters, keys should not matter. I assume that this behaviour could even be beneficial as a default because it significantly reduces the size.

Another minor issue I noticed is that uPickle uses the full class name, i.e., it includes the package names. As a workaround I could just rewrite the aeson JSON tree that I'm parsing (as I do with ``t`` and ``c`` right now), but setting a relative path could further shorten the JSON.

It may make sense to create a ``Configuration`` trait that contains these kind of settings.

lihaoyi:
Yeah it used to be like that. Then I put the keys in because everyone else was complaining that it didn't interoperate well with all their existing JSON infrastructure =P 

You should be able to shadow the macroR and macroW pretty easily to make this happen. In fact, it probably doesn't even need to be a macro anymore: just write the 10 or 22 overrides for differently sized `ProductN` traits and you're done.

tindzk:
Ok, that might work as a temporary solution. Yet, I'd like to make uPickle more configurable, so that I can easily use different set-ups within the same project. This is solved nicely in aeson with a configuration object. As described in #55, I did try something similar for uPickle, but ran into problems with implicits. Perhaps you can think of a better way to implement it — without implicits?

lihaoyi:
Your choice is between implicits or cakes. I've had great success using cakes for Scalatags, where the entire library is essentially giant cake that I instantiate 2-3 times for Text/DOM/React versions. Perhaps you could try doing the same with uPickle? The advantage of cakes is, of course, they don't use implicit resolution at all and thus are much less fragile to e.g. bad inference or imports. 

[...]

PR #65, opened 2015-01-24
Extracted annotators to make them configurable

### LINKED ISSUE #64: Customizable discrimination annotations
It would be nice if annotation format could be customized, for example, in order to have 
```
{ "$variant": "MyClass", "key": "value" }
````
instead of 
```
["your.package.MyClass", {"key": "value"}]
```

While such approach can lead to collisions, with some fancy discriminator key it will be very unlikely (and it is not really a problem if it can be configured). It would also be nice to be able to store only the last part of the discriminator value (or, for example, two parts if `MyClass` is defined in an `object`).

As far as I can see (I may be wrong, though) it should not be very difficult - currently the autogenerated picklers depend on `Internal.annotate` behavior; I think it should be possible to extract the annotation behavior to another implicit and provide the default one with the current behavior.

Having the discriminator as a part of the object is important (at least for our project) for migration reasons (a lot of other serializators are using this format) and for working with the serialized objects - for example, mongodb queries tend to be simpler without arrays.

### PR DESCRIPTION
**Pull request by netvl**
This fixes #64 

Default behavior is not changed, so this PR is backwards-compatible. In order to override the annotation strategy an `Annotator` instance should be brought into scope:

```scala
package example

import upickle._
import upickle.annotator.key._

sealed trait Root
case class A(value: Int) extends Root

write(A(10))  // {"$variant": "example.A", "value": 10}
```

It is also possible to override the discriminator key:
```scala
package example

import upickle._

implicit val customKeyAnnotator = annotator.key.keyAnnotator("$custom$")

sealed trait Root
case class A(value: Int) extends Root

write(A(10))  // {"$custom$": "example.A", "value": 10}
```

If one wants to override the default annotator, the corresponding implicit has to be present in the scope where `Reader`s/`Writer`s are defined.


lihaoyi:
- I don't like default implicits in this case; too easy to accidentally leave out an import and have two parts of your program not-talking to each other. This was a problem earlier with the macro default RW accidentally being used to RW standard-library stuff for which there's already a RW in Implicits, before I made them imported by default, and it's probably still quite dangerous.I don't mind forcing people to do an `import upickle.config.default` at the top of each file if they don't customize it
- I'd prefer making your variant-encoding (using `$variant`) the `upickle.config.default`; we can leave the current behavior as `upickle.config.old`. This will involve changing most of the tests, but that should be straightforward.
- @tindzk this would be of interest to you in https://github.com/lihaoyi/upickle/issues/55. Do you want to try out @netvl's branch? He's got a pretty clever way of modifying how the case class serialization is done, and you could easily modify it to control the serialization as-a-whole and e.g. get your case-class-as-a-tuple behavior

netvl:
I agree, explicit configuration decisions are better; I just wanted to keep the backwards compatibility. If it is not a problem, then I'll remote the default.

> import upickle.config.default

Do you want some kind of general configuration? I'm asking this because I'm currently working on the customization of case class fields serialization mentioned in #43 (that `Filter` stuff). I don't have a clean design yet, but I guess it should be integrated with common configuration if there will be one.

lihaoyi:
Backwards compatibility would be nice, but as a pre-`1.0` library, I think we can make a few changes without people getting too upset ^_^

Yeah a general configuration mechanism would be ideal. @tindzk would like one as well. How general is open for debate: maybe someday we'll be serializing to protobuffers and other things, who knows! For now just customizing case-classes should be sufficient I guess

netvl:
(BTW, serialization to something else than `Js` tree, e.g. some foreign BSON ADT, would be a very nice thing to have in our project as well)

Ok, I guess for now I indeed can make `upickle.config.default` and `upickle.config.old` objects which for now will only contain annotators.

tindzk:
I'm glad you did something similar to what I tried for #59 (strangely, I couldn't get it working with ``implicitly``). Indeed, we should think of a general configuration method. @lihaoyi suggested to have a look at his branch https://github.com/lihaoyi/upickle/tree/cake. Once I'm back from my trip I will check whether the Cake approach could lead to a cleaner solution than implicits.

netvl:
@tindzk, I think it works because I don't pass the configuration object through macros - I just generate pickler code with `implicitly[Annotator]` which is resolved after the macro expansion, not during it.

I still need to wrap my head around the cake approach (I know what cake pattern is, but its usage in this context is still vague to me), but anyway implicits are lightweight and don't need refactoring. And when properly done they still allow cake-like composition of configuration.

netvl:
@lihaoyi, I've refactored annotators classes arrangement and added the configuration objects with implicits. Now the desired configuration has to be imported explicitly.

I think I also have to update the readme but I'll do it later in a separate commit, probably in this pull request as well.

netvl:
@lihaoyi, hmm, I'm now not sure if the readme has to be updated now. It holds instructions for 0.2.5 which is the current stable version. I think it should be updated before the new release.

Other than that, this PR is ready for merge.

js/shared/main/scala/upickle/Annotator.scala
```diff
@@ -0,0 +1,43 @@
...
+ */
+@implicitNotFound("Couldn't find any Annotator implementation in scope. Make sure you have imported some configuration, e.g. `import upickle.config.default._`")
+trait Annotator {
+  def annotate[V](r: Reader[V], name: String): Reader[V]
+  def annotate[V: ClassTag](w: Writer[V], name: String): Writer[V]
+}
+
+object Annotator {
+  // It drops down to array annotations if the serialized object is not represented as a `Js.Obj`
+  def keyAnnotator(key: String): Annotator = new Annotator {
```
tindzk: Would it be possible to use a ``case class`` here?

netvl: I'm not sure I understand what you mean :)

tindzk: Ok, let me rephrase: Is there a specific reason why you didn't write: ``case class KeyAnnotator(key: String) extends Annotator { ... }``?

netvl: Thanks for the clarification.

Well, case classes are semantically intended to be a data transfer objects - they are needed to keep and pass any data around. Annotator is not a data object, and `key` is just a parameter for its functionality. Hence it is not a case class.

netvl:
BTW, just for the information, using annotators now looks like this:
```scala
package example

import upickle._
import upickle.config.default._

sealed trait Root
case class A(value: Int) extends Root

write(A(10))  // {"$variant": "example.A", "value": 10}
```

With a custom key:
```scala
package example

import upickle._

object customConfig extends upickle.config.DefaultConfig {
  override implicit val annotator = Annotator.keyAnnotator("$custom$")
}
import customConfig._

sealed trait Root
case class A(value: Int) extends Root

write(A(10))  // {"$custom$": "example.A", "value": 10}
```

It allows writing custom configuration easily and leaves room for more configuration options.

netvl:
@lihaoyi, I'm very sorry to bother you, but do you plan to merge this? We would like to use uPickle in our project, but without this change we can't :(

lihaoyi:
@netvl yeah I do, but unfortunately, I have work/other things going on and haven't had time to review =/ I'll get to it as soon as I can. Perhaps publishing your own fork for the time being may be a reasonable workaround.

The hardships of free-time development...

netvl:
Yeah, that's what I did for now. Anyway, thank you very much for the library :)

knshiro:
@lihaoyi 
Thank you very much for your library
I would love to use this in [sbt-packer](https://github.com/en-japan/sbt-packer), is there any chance for a merge?

lihaoyi:
Rendered redundant with http://lihaoyi.github.io/upickle-pprint/upickle/#CustomConfiguration, not exactly the same but pretty close, including trivial support for all the examples provided here and more, e.g. the very-very-old case-classes-as-tuple serialization! @knshiro @netvl @tindzk FYI

Issue #68, opened 2015-02-12
Serialization macros intermittently fail


tealsoftware:
I have a relatively simple case class graph to serialize, with one instance of 'sealed trait' implemented by multiple case classes.

It fails when compiled with my application, as described below. However, when I remove all other code and only compile the case classes and Json.scala, everything works fine! 

The thing needed to make it work is to remove the code (Scala Android app) not related to upickle in any way!

When I compile the whole code through sbt, I get very many messages such as:
```
The referenced trait [[Answer]] does not have any sub-classes. This may happen due to a limitation of scalac (SI-7046) given that the trait is not in the same package. If this is the case, the hierarchy may be defined using integer constants.
```

and then a few compile messages such as
```
[error] C:\projects\myproject\android\src\main\scala\package\SynchronizationActivity.scala:28: exception during macro expansion:
[error] java.lang.AssertionError: assertion failed
[error]         at scala.Predef$.assert(Predef.scala:151)
[error]         at upickle.Macros$.macroWImpl(Macros.scala:49)
[error]         at sun.reflect.GeneratedMethodAccessor3.invoke(Unknown Source)
[error]         at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[error]         at java.lang.reflect.Method.invoke(Method.java:606)
[error]         at scala.reflect.macros.runtime.JavaReflectionRuntimes$JavaReflectionResolvers$$anonfun$resolveJavaReflectionRuntime$2.apply(JavaReflectionRuntimes.scala:34)
[error]         at scala.reflect.macros.runtime.JavaReflectionRuntimes$JavaReflectionResolvers$$anonfun$resolveJavaReflectionRuntime$2.apply(JavaReflectionRuntimes.scala:22)
[error]         at scala.tools.nsc.typechecker.Macros$class.macroExpandWithRuntime(Macros.scala:755)
[error]     info(s"This would be sent to server: ${Serialize(qr)}")
[error]                                                     ^
[error] 15 errors found
[error] (compile:compile) Compilation failed
```

The trait it has a problem with is roughly
```scala
sealed trait Answer 
case class AnswerWithDefects(defects: Option[Seq[Defect]] = None) extends Answer
case class AnswerYesNo(expectedAnswer: Boolean, answer: Option[Boolean] = None) extends Answer 
```

The code doing the serialization is 
```scala
package mypackage.json
import upickle._

object Serialize {
[...]

Issue #31, opened 2014-09-29
Macro implicits don't work outside package


japgolly:
Hey. Check this out:
https://github.com/japgolly/upickle/commit/9a976af73ed46eeb7364829e457a973719ed9a7b

`Eg2` compiles but `Eg1` doesn't.

```
[error] upickle/jvm/shared/test/scala/upickle/blah/blah.scala:7: uPickle does not know how to read [upickle.elsewhere.YYY]s; define an implicit Reader[upickle.elsewhere.YYY] to teach it how
[error]   def blah(j: String): YYY = read[YYY](j)
[error]                                       ^
[warn] one warning found
[error] one error found
[error] upickle/js/shared/test/scala/upickle/blah/blah.scala:7: uPickle does not know how to read [upickle.elsewhere.YYY]s; define an implicit Reader[upickle.elsewhere.YYY] to teach it how
[error]   def blah(j: String): YYY = read[YYY](j)
[error]                                       ^
[warn] one warning found
[error] one error found
```

tindzk:
After I made changes in some unrelated areas of my code, this issue suddenly showed up.

@l15k4 Could you give an example how inheriting from ``Implicits`` fixed the problem? It contains unimplemented functions, so how would that work?

I also defined an implicit in my ``Protocol`` class which is in a different package. Then, I evaluate this implicit where ``upickle.write()`` is called, but to no avail.

l15k4:
@tindzk I can't figure out what I meant by "I default to either inheriting from Implicits trait". To my knowledge if you have custom R/W implicits in a different package than you need to directly `import` them to the scope where `upickle.write()` is called and it should work.

tindzk:
There must be something in Widok that collides with upickle. When I comment out ``import org.widok._``, the errors vanish. My first idea was that this is related to the implicits that I define [here](https://github.com/widok/widok/blob/master/src/main/scala/org/widok.scala), but even if I import these manually, I cannot reproduce this issue. I wish scalac would provide a bit more guidance. So a temporary solution for now is to encapsulate any upickle-related code in a separate file and ensure that no conflicting imports are made.

l15k4:
[...]

PR #69, opened 2015-02-15
Reading an invalid object should not throw any exceptions

tindzk:
This currently prevents us from easily defining our own composite
readers, which is often necessary; see #31 and #68. Without the
exceptions, ``implicitly[Reader[_]].read`` calls can be chained using
``orElse()``, as done internally in uPickle.

```scala
sealed trait Base
case class Child1() extends Base
case class Child2() extends Base

import upickle._
implicit val BaseR: Reader[Base] = Reader[Base](
  implicitly[Reader[Child1]].read
    .orElse(implicitly[Reader[Child2]].read)
)
```

The exceptions are only useful for debugging purposes.


lihaoyi:
Looks reasonable. In future it may be worth converting the `PartialFunction[T, R]` reads into a `T => Either[Error, R]` so we can preserve both the rich-error-data as well as the chainability

PR #73, opened 2015-03-04
Implement UUID support

k3d3:
As the title says - support for reading and writing UUIDs by default. Simply encode them as strings in UUID format.


lihaoyi:
Looks good to me, thanks!

lihaoyi:
Published as 0.2.8

PR #77, opened 2015-03-11
make macro expansion errors more helpful

romansky:


tindzk:
The code is not indented properly. Have you tried using ``c.abort()`` instead of throwing an exception?

romansky:
Right, fixed.

lihaoyi:
Not exactly much more helpful, but I suppose better than an `AssertionError`. Thanks!

PR #78, opened 2015-03-23
README: Remove Bintray resolver

tindzk:
As uPickle is pushed to Maven Central, the Bintray resolver is not needed anymore.


lihaoyi:
Actually it's `jawn` that's now on maven central, but yeah we no longer need @non's resolver

PR #86, opened 2015-06-02
Reading unquoted numbers as longs

senia-psm:
https://github.com/lihaoyi/upickle/issues/66


lihaoyi:
I'm gonna not-merge this since it's trivial to set this up in your own project. It introduces an asymmetry between reading and writing I'm not comfortable with including in mainline by default

lihaoyi:
Just C&P that exact snippet into your own object and import it where you want this behavior =)

senia-psm:
@lihaoyi, you already have this kind of asymmetry in all `NumericReadWriter`: read `Str` and `Num`, write `Num`.

lihaoyi:
AFAICT that writes `Num`s and `Str`s too because `Infinity` and `-Infinity` cannot be represented as JSON numbers

PR #91, opened 2015-07-08
documentation link was broke

mrmechko:


lihaoyi:
Seems I already fixed it independently; thanks for the PR though!

PR #109, opened 2015-08-28
Small improvement for standard usecase

LubomirVarga:
Just used trait in read command instead of concrete class. It is more common use case.


lihaoyi:
Bug bankruptcy

PR #113, opened 2015-10-07
Add support for `BigInt` and `BigDecimal`

guersam:
Cross source code seems a bit like an overkill but it clearly shows what to do when we drop Scala 2.10 support.


lihaoyi:
Bug bankruptcy

PR #114, opened 2015-10-08
Js.Obj reader and writer

prettynatty:


lihaoyi:
This doesn't really do the right thing for the various JSON structures you might try to use it on

```scala
scala> write(upickle.Js.Obj())
res2: String = {}

scala> write(upickle.Js.Num(1))
res3: String = {"$type":"upickle.Js.Num","value":1}
```

lihaoyi:
@prettynatty want to complete this with picklers and unpicklers for `Val` that would work for everything?

lihaoyi:
Closing due to inactivity

PR #122, opened 2015-11-28
enable non serializable objects. 

bblfish:
This is needed for turning JS CryptoAPI opaque objects into an AST to send to IndexedDB as explained in https://groups.google.com/d/msg/scala-js/9QLzClZSGtQ/1pJ8fPbxCAAJ
For a usage see the "non hacky Key saving to DB" commit on read-write-web below, which allows me to remove the hack I needed for https://github.com/viagraphs/scalajs-rx-idb/issues/3

### DIFF
--- a/upickle/js/src/main/scala/upickle/json/package.scala
+++ b/upickle/js/src/main/scala/upickle/json/package.scala
@@ -1,6 +1,8 @@
 package upickle
-import acyclic.file
-import scalajs.js
+
+import upickle.Js.Other
+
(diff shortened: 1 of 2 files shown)


bblfish:
the reading works as is, but not the writing, so there is something else that needs to be changed... ( looking )

bblfish:
This will not work without a much deeper rewrite. I wrote up the details in https://github.com/lihaoyi/upickle-pprint/issues/127

PR #125, opened 2015-12-12
Added support for field names in PPrint

patrick-premont:
The motivation for this feature is to obtain a more verbose and descriptive representation of objects by also displaying field names. This can be especially useful when logging. Although JSON could be used for that, it is bit more noisy and a less natural description of Scala objects. Also this is legal Scala syntax so objects printed in such a manner could still be parsed and evaluated by Scala. 

```
scala> import pprint.Config.Colors._
import pprint.Config.Colors._

scala> case class Test(name: String, value: Int)
defined class Test

scala> pprint.pprintln(Test("a" * 10, 1))
Test("aaaaaaaaaa", 1)

scala> pprint.pprintln(Test("a" * 10, 1))(implicitly, PPrintConfig.copy(showFieldNames = true))
Test(name = "aaaaaaaaaa", value = 1)

scala> pprint.pprintln(Test("a" * 80, 1))(implicitly, PPrintConfig.copy(showFieldNames = true))
Test(
  name = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
  value = 1
)
```

This is a bit preliminary in that I have added only one test. I can't think if there are any corner cases to consider. Also Scala macros aren't my strong suit. However the changes to the code are very minimal. Looking forward to your feedback.

For full disclosure, I do see one respect in which this change might not be enough for my needs long term: It might be nice in the future to support an annotation on some fields whose name should be omitted in spite `showFieldNames = true`. I already see a few cases where I'd like that feature. Your renaming mechanism could perhaps be extended to address similar issues with field names (Option.x is not very useful). My feeling is that we do not need to address the support for such exceptions as part of this PR; what do you think ?

While reviewing your PR guidelines (those for Ammonite), I notice I haven't added to the documentation. I should do that.

Thanks

### DIFF
--- a/pprint/shared/src/main/scala/pprint/PPrint.scala
+++ b/pprint/shared/src/main/scala/pprint/PPrint.scala
@@ -108,9 +108,9 @@ trait PPrinter[-A] {
 
 object PPrinter extends LowPriPPrinter{
   // Things being injected into PPrinterGen to keep it acyclic
-  
+
   def apply[T](r: (T, Config) => Iter[String]): PPrinter[T] = {
-    new PPrinter[T]{ 
+    new PPrinter[T]{
       def render0(t: T, c: Config) = {
         if(c.height > 0)
           takeFirstLines(c, r(t, c))
@@ -447,9 +447,13 @@ object Internals {
               q"""
               $pkg.PPrint[$targetType](
                 $pkg.PPrinter.ChunkedRepr[$targetType](
-                  $pkg.Chunker[$targetType]( ($t: $targetType, $cfg: $pkg.Config) =>
-                    $pkg.Chunker.$tupleChunker[..$argTypes].chunk($w, $cfg)
-                  )
+                  $pkg.Chunker[$targetType]{ ($t: $targetType, $cfg: $pkg.Config) =>
+                    val chunkedArgs = $pkg.Chunker.$tupleChunker[..$argTypes].chunk($w, $cfg)
+                    if ($cfg.showFieldNames) chunkedArgs
+                      .zip(Iterator.apply(..${argSyms.map(_.name.toString + " = ")}))
+                      .map{case (chunkedArg, field) => Iterator.single(field) ++ chunkedArg}
+                    else chunkedArgs
+                  }
                 )
               )
               """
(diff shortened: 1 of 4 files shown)


patrick-premont:
Hmm... There's a compile problem under 2.10.4. Let me know if you have an idea about that.

stewSquared:
@lihaoyi If this feature looks worthwhile, I'd be happy to update this pull request and add documentation.

pprint/shared/src/main/scala/pprint/PPrint.scala
```diff
@@ -447,9 +447,13 @@ object Internals {
...
                 $pkg.PPrinter.ChunkedRepr[$targetType](
-                  $pkg.Chunker[$targetType]( ($t: $targetType, $cfg: $pkg.Config) =>
-                    $pkg.Chunker.$tupleChunker[..$argTypes].chunk($w, $cfg)
-                  )
+                  $pkg.Chunker[$targetType]{ ($t: $targetType, $cfg: $pkg.Config) =>
+                    val chunkedArgs = $pkg.Chunker.$tupleChunker[..$argTypes].chunk($w, $cfg)
+                    if ($cfg.showFieldNames) chunkedArgs
+                      .zip(Iterator.apply(..${argSyms.map(_.name.toString + " = ")}))
+                      .map{case (chunkedArg, field) => Iterator.single(field) ++ chunkedArg}
+                    else chunkedArgs
```
lihaoyi: Can you extract this into a function? In general the code we generate inside macros should be minimal and any non-trivial logic like this stuff should end up in a helper function somewhere.

lihaoyi:
Sorry for the delay! In principle I think this is a good idea, if someone wants to pick it up and bring it up to date I'll merge/publish it

dportabella:
what is the status of this feature implementation? i see that there is a compilation problem on scala 2.10. is it a simple thing to correct, or it is a major problem?

lihaoyi:
Bug bankruptcy

PR #126, opened 2015-12-19
Fixes #104

### LINKED ISSUE #104: Derivation doesn't work for hierarchies starting from `sealed abstract class`
I observed this in relation with pprint which doesn't work (i.e. falls back to `toString`) for `Either[X, Y]` values because `Either` is defined as an `sealed abstract class`.

Voltir:
Abstract classes are treated like traits as far as Derive is concerned.

### DIFF
--- a/test/Main.scala
+++ b/test/Main.scala
@@ -2,6 +2,6 @@
 object Main{
   def main(args: Array[String]): Unit = {
     val x = derive.PythonAst
-    implicitly[upickle.default.Writer[derive.PythonAst.expr]]
+    //implicitly[upickle.default.Writer[derive.PythonAst.expr]]
   }
 }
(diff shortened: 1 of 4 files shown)


test/Main.scala
```diff
@@ -2,6 +2,6 @@
 object Main{
   def main(args: Array[String]): Unit = {
     val x = derive.PythonAst
-    implicitly[upickle.default.Writer[derive.PythonAst.expr]]
+    //implicitly[upickle.default.Writer[derive.PythonAst.expr]]
```
lihaoyi: What's this though?

lihaoyi:
Looks fine, except for that commented out thing

Voltir:
That appears to be the source of the build instability I was talking about on gitter. Uncommenting that line and then running clean compile in sbt would fail with a compile time error every half dozen attempts or so.

lihaoyi:
This looks good. I'll look into the PythonAst stuff offline.

Thanks!

Issue #124, opened 2015-12-04
Derive.TypeKey lacks hashCode implementation


ghik:
`Derive.TypeKey` has `equals` overridden but does not properly override `hashCode`. This can lead to patologies like:

```scala
val tk1 = TypeKey(typeOf[String])
val tk2 = TypeKey(typeOf[java.lang.String])
tk1 == tk2 // true
tk1.hashCode == tk2.hashCode // false!!!
```

Voltir:
I looked at this a bit and realized that I'm not actually sure what a proper hashcode would be.

lihaoyi:
This has happened, more or less

ghik:
I also saw shapeless implement `Type`-keyed maps. There was no `hashCode` implemented at all, but the consequences were avoided by using `ListMap` instead of `HashMap` :)

PR #128, opened 2016-01-22
fixes an issue impacting derivation of sealed traits

Voltir:
I think #124 was the root cause of this particular problem. The problem was that during the macro derivation the `seen` set of TypeKeys was returning false in instances where it should have returned true and thus was generating "ambiguous" implicits for the same type by generating multiple instances for that type. 

In the test code's case, the problematic instance was `shared.common.Message`. Giving an explicit, and hopefully sensible, hashCode fixes the problem as `seen` returns the expected result.


lihaoyi:
image

PR #129, opened 2016-01-28
Allow custom handling of nulls

bwbecker:
null-handling:  This addresses https://github.com/lihaoyi/upickle-pprint/issues/63 which says:

> Currently the way Reader/Writer handle nulls is hardcoded: null is always converted to and from Js.Null, and corresponding methods are final so they can't be overridden. However, sometimes it makes sense to handle nulls differently, for example, when overriding [de]serialization of options.

> Is it possible to allow custom handling of nulls?


l15k4:
Is this supposed to fix the problem that : 
```scala
  implicit def someToObject[T : Writer]: Writer[Some[T]] = Writer[Some[T]] {
    case Some(v) =>
      writeJs[T](v)
    case x =>
      upickle.Js.Null
  }
```

doesn't get applied? For some reason I cannot override the default Option implicit writer -> it is being serialized as array

bwbecker:
Yes.  You can’t override the Null because it’s picked off very early in the process in a final method.

Byron

> On Feb 23, 2016, at 8:48 AM, Jakub Liska <notifications@github.com> wrote:
> 
> Is this supposed to fix the problem that :
> 
>   implicit def someToObject[T : Writer]: Writer[Some[T]] = Writer[Some[T]] {
>     case Some(v) =>
>       writeJs[T](v)
>     case x =>
>       upickle.Js.Null
>   }
> doesn't get applied? For some reason I cannot override the default Option implicit writer -> it is being serialized as array
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/lihaoyi/upickle-pprint/pull/129#issuecomment-187705942>.
> 


---------------------------------------------
Byron Weber Becker
School of Computer Science
University of Waterloo, Waterloo, Ontario  N2L 3G1
Email: bwbecker@uwaterloo.ca
Phone:  519-888-4567 x34661
Fax:  519-888-4380 (include my email on cover page)

Advising FAQ:  www.cs.uwaterloo.ca/current/faq

upickle/shared/src/main/scala/upickle/Types.scala
```diff
@@ -90,9 +90,31 @@ trait Types{ types =>
   trait Reader[T]{
     def read0: PF[Js.Value, T]
 
-    final def read : PF[Js.Value, T] = ({
+    private def readNull: PF[Js.Value, T] = {
       case Js.Null => null.asInstanceOf[T]
-    }: PF[Js.Value, T]) orElse read0
+    }
+
+    private def invalid: PF[Js.Value, T] = {
```
lihaoyi: This is un-used

upickle/shared/src/main/scala/upickle/Types.scala
```diff
@@ -90,9 +90,31 @@ trait Types{ types =>
...
+     * 
+     * This reverses the order.  However, in the original ordering read0 will throw
+     * an exception when called with a value not in the domain.  In the new ordering
+     * it won't throw that exception because isDefinedAt is called and prevents
+     * the call to apply.  By adding it again at the end, the original behaviour of
+     * throwing the exception is restored.
+     */
+
+    //final def read: PF[Js.Value, T] = readNull orElse read0
+    final def read: PF[Js.Value, T] = read0 orElse readNull orElse read0
```
lihaoyi: Why are we doing `read0 orElse readNull orElse read0`?

upickle/shared/src/main/scala/upickle/Types.scala
```diff
@@ -90,9 +90,31 @@ trait Types{ types =>
...
+     * a string).
+     * 
+     * This reverses the order.  However, in the original ordering read0 will throw
+     * an exception when called with a value not in the domain.  In the new ordering
+     * it won't throw that exception because isDefinedAt is called and prevents
+     * the call to apply.  By adding it again at the end, the original behaviour of
+     * throwing the exception is restored.
+     */
+
+    //final def read: PF[Js.Value, T] = readNull orElse read0
```
lihaoyi: Shouldn't leave commented out code lying around

upickle/shared/src/main/scala/upickle/Types.scala
```diff
@@ -90,9 +90,31 @@ trait Types{ types =>
...
+    private def readNull: PF[Js.Value, T] = {
       case Js.Null => null.asInstanceOf[T]
-    }: PF[Js.Value, T]) orElse read0
+    }
+
+    private def invalid: PF[Js.Value, T] = {
+      case any => throw Invalid.Data(any, "name")
+    }
+
+    /*
```
lihaoyi: This huge comment block should be in the pull request text, not embedded in the codebase

upickle/shared/src/test/scala/example/OptionsAsNullTests.scala
```diff
@@ -0,0 +1,82 @@
...
+import acyclic.file
+import upickle.{Js, TestUtil}
+import utest._
+
+import example.Simple.Thing
+
+  case class Opt(a:Option[String], b:Option[Int])
+
+
+object OptionsAsNullTests extends TestSuite{
```
lihaoyi: I'm not sure why no tests caught the fact that `invalid` and `read` as defined above are probably wrong. Could you add a regression test to ensure that sort of mistake doesn't happen in future?

lihaoyi:
Could you add this to the docs?

lihaoyi:
Other than the comments here it looks good to land, just need to sort out the things i mentioned and i'll merge/publish it

bwbecker:
Thanks for the comments.

I know "read0 orElse readNull orElse read0" looks funny but without that last "orElse read0" tests already in the test suite fail.  Including it, they pass.  I'd like to replace it with "orElse invalid", but don't know how to get the name to pass to the exception.

The big comment you objected to is trying to explain why that last "orElse read0" is there.  It's weird enough that I'd leave some sort of comment behind to explain, but maybe much reduced.

Can you say a bit more about what you'd like added to the documentation and where?

lihaoyi:
documentation should go in to the section under [Custom Picklers](http://www.lihaoyi.com/upickle-pprint/upickle/#CustomPicklers)

I've read this several times and still have no idea what the `orElse read0` is for. Do you have a concrete problem you want to solve that I can help with? I don't really know what you're trying to do that you don't know

bwbecker:
What I'm trying to accomplish is illustrated by the tests in test/scala/example/OptionsAsNull.  I'd like to redefine the handling of Option[T] as follows:

```scala
      object OptionPickle extends upickle.AttributeTagged{

        override def OptionW[T: Writer]: Writer[Option[T]] = Writer{
          case None => Js.Null
          case Some(s) => implicitly[Writer[T]].write(s)
        }

        override def OptionR[T: Reader]: Reader[Option[T]] = Reader {
          case Js.Null => None
          case v:Js.Value => Some(implicitly[Reader[T]].read.apply(v))
        }
      }
```
The application is to turn json generated from a Postgres database into Scala data structures.
The above code does not work with the existing uPickle because Reader[T] in Types.scala picks off Js.Null and turns it into null rather than None.  It does this before my code (now in ```read0```) is called:

```scala
    final def read: PF[Js.Value, T] = readNull orElse read0
```
The method is final, so I don't think there is anything I can do about it other than change the library itself.
My first attempt was the very obvious
```scala
    final def read: PF[Js.Value, T] = read0 orElse readNull
```
However, this causes a number of existing tests to fail.  An example is
```scala
        intercept[Invalid.Data] { read[Boolean]("\"lol\"") }
```
I believe the ```read[Boolean]``` eventually executes
```scala
  private[this] val booleanReaderFunc: JPF[Boolean] = Internal.validate("Boolean"){
    case Js.True => true
    case Js.False => false
  }
```
```booleanReaderFunc``` becomes ```read0``` in ```readNull orElse read0``` and throws an error when the input is not Js.True or Js.False.  Switching the order to ```read0 orElse readNull``` causes a different error to be thrown.  But adding read0 back at the end (```read0 orElse readNull orElse read0```) causes the original error to be thrown again.  
I'd prefer to write something like ```read0 orElse readNull orElse InvalidData(myValue, name)```  but I don't have all the needed information for the exception.

**Comment by bwbecker** [comment]
I've updated my branch as follows:

* Cleaned up the Reader trait in Types.scala (removing comment block, unused method, etc)
* Revised my test suite.
* Added documentation.

I replaced the implementation of ```read``` is ```Reader``` with:

```scala
    final def read: PF[Js.Value, T] = new PartialFunction[Js.Value, T] {
      def isDefinedAt(x: Js.Value) = read0.isDefinedAt(x)
      def apply(v1: Js.Value): T = read0.applyOrElse(v1, readNull orElse read0)
    }

```
In a more verbose form, it's equivalent to 
```scala
    final def read: PF[Js.Value, T] = new PartialFunction[Js.Value, T] {
      def isDefinedAt(x: Js.Value) = read0.isDefinedAt(x)
      def apply(v1: Js.Value): T = if (read0.isDefinedAt(v1)) {
        read0.apply(v1)
      } else {
        readNull.applyOrElse(v1, read0)   // old behaviour
      }
    }
```
But this is also equivalent to my original solution:
```scala
    final def read: PF[Js.Value, T] = read0 orElse readNull orElse read0
```

Another approach to the problem is to take handling of nulls out of the Reader trait completely.  This would involve changes to many of the builtin readers so they they handled nulls themselves.  It could also break code that others have written that relies on the original behaviour.  

Of the four proposed solutions, I prefer my original ```read0 orElse readNull orElse read0```.

bwbecker:
The merge conflict is the version number in Constants.scala.

lihaoyi:
Ok I understand the issue at hand here: `readNull` throws a different exception on match-not-found than the custom `read0`. By re-ordering it, we expose the `readNull` exception, and you were trying to force `read0` to be called at the end so *it's* exception gets thrown.

I am satisfied with the final result you came up with. Thanks for pushing this through!

bwbecker:
Thank you for sticking with it and accepting the PR.


> On Apr 23, 2016, at 11:59 AM, Li Haoyi <notifications@github.com> wrote:
> 
> Ok I understand the issue at hand here: readNull throws a different exception on match-not-found than the custom read0. By re-ordering it, we expose the readNull exception, and you were trying to force read0 to be called at the end so it's exception gets thrown.
> 
> I am satisfied with the final result you came up with. Thanks for pushing this through!
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub <https://github.com/lihaoyi/upickle-pprint/pull/129#issuecomment-213771121>


---------------------------------------------
Byron Weber Becker
School of Computer Science
University of Waterloo, Waterloo, Ontario  N2L 3G1
Email: bwbecker@uwaterloo.ca
Phone:  519-888-4567 x34661
Fax:  519-888-4380 (include my email on cover page)

Advising FAQ:  www.cs.uwaterloo.ca/current/faq

PR #146, opened 2016-04-04
Made derive.key a case class

### LINKED ISSUE #144: Make derive.key a case class such that shapeless.Annotation can pick it up
Shapeless has this interesting `Annotation[A, T]`. From the comments:

> If type `T` has an annotation of type `A`, then an implicit `Annotation[A, T]` can be found, and its `apply` method provides an instance of the annotation.

A current limitation is that such an implicit is found only when the annotation is a case class. It would thus be useful to make `derive.key` a case class such that other mechanisms could utilise `shapeless.Annotation` to extract its values.

jppellet:
This fixes #144.

### DIFF
--- a/derive/shared/src/main/scala/derive/Derive.scala
+++ b/derive/shared/src/main/scala/derive/Derive.scala
@@ -10,7 +10,7 @@ import language.higherKinds
  * to use a custom string as the key for that class/field rather than the
  * default string which is the full-name of that class/field.
  */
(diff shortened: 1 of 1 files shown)


lihaoyi:
No idea why this isn't auto-closed but I merged it in https://github.com/lihaoyi/upickle-pprint/commit/e36ab845adbb7c92c2aa24732c8497404ef8a06a

PR #149, opened 2016-04-13
Cached implicit

Voltir:
This PR introduces shaplesses version of implicit resolution as well as the "macroRW" macro to create a 'Reader[T] with Writer[T]` in a single statement.


lihaoyi:
lol your test cases just passed in CI without any of your `Derive.scala` changes...

lihaoyi:
@Voltir can you add some more tests to https://github.com/lihaoyi/upickle-pprint/compare/test that would fail without your changes to `Derive.scala`?

lihaoyi:
nm i found a case...

lihaoyi:
Should be fixed by https://github.com/lihaoyi/upickle-pprint/commit/37f7ee7d4975afaf1f6e1b7fcab6677c0a10cb73

Issue #143, opened 2016-04-01
derive calls writeJs and hard-codes a dependency on upickle


jppellet:
When wrapping the default value of case class arguments in JSON values, `derive` does this:

```scala
        val defaults = argSyms.zipWithIndex.map { case (s, i) =>
          val defaultName = newTermName("apply$default$" + (i + 1))
          companion.tpe.member(defaultName) match {
            case NoSymbol => q"null"
            case _ => q"${c.prefix}.writeJs($companion.$defaultName)"
          }
        }
```
(https://github.com/lihaoyi/upickle-pprint/blob/master/derive/shared/src/main/scala/derive/Derive.scala#L317)

Hard-coding the call to `writeJs` means that `derive` won't derive type class instances if this method doesn't exist.

lihaoyi:
lol

not sure what else to say, we should fix it :stuck_out_tongue: :stuck_out_tongue: :stuck_out_tongue:

PR #152, opened 2016-04-21
Fix 143

Voltir:
The fix for #143


lihaoyi:
I have no idea but the test pass so ok

PR #171, opened 2016-09-03
implicits for pickling Js.Value

fdietze:


fdietze:
I found an example which does not work:

```scala
package example

import upickle.Js

case class A(bs: List[B] = Nil)
case class B(nums: List[Js.Num] = Nil)

object Main {
  upickle.default.write(A())
}
```

error:
```
Example.scala:9: type mismatch;
 found   : upickle.default.Writer[(example.A,)]
 required: upickle.default.Aliases.W[(List[example.B],)]
    (which expands to)  upickle.default.Writer[(List[example.B],)]
Error occurred in an application involving default arguments.
  upickle.default.write(A())
```

Any Ideas?

lihaoyi:
The error you're showing seems to be a independent issue - not sure what it is, but it should be unrelated to the `Js.Value` picklers you're introducing here. I'll merge this first and later maybe we can go figure that out

PR #175, opened 2016-10-06
Write long number that have no decimals without decimals

naksu:
The benefit is that millisecond timestamps are not encoded with scientific notation.

For example:
  1475741505173L
  becomes 1475741505173
  instead of 1.475741505173E12

I'm getting a lot of warnings in sbt (deprecation, scala version). Is that normal?


lihaoyi:
This looks great

Yeah the project is full of warnings =/ someday i'll clean it up

PR #187, opened 2016-11-29
Case classes can now also render their field name.

stephanh:
Whether the field name should be shown is configured via the `showFieldName` flag in the Config.

Rather than change the Tuple pretty printer I changed the code to derive a targeted pretty printer for each case class.

### DIFF
--- a/pprint/shared/src/main/scala/pprint/Core.scala
+++ b/pprint/shared/src/main/scala/pprint/Core.scala
@@ -14,13 +14,16 @@ import acyclic.file
  * @param renames A map used to rename things to more common names, e.g.
  *                renamig `WrappedArray` to `Array` or getting rid of
  *                TupleN *
+ * @param showFieldName Should the field name of a case class be displayed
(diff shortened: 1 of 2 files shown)


stephanh:
I only saw #125 after I raised this PR.

This seem to compile and pass all tests under 2.10.6 on my machine.

There are a bunch of failing tests on the master branch for 2.11.8. It still compiles and the number of passing/failing tests hasn't changed.

dportabella:
this feature is a must. :)
what is the status?

lihaoyi:
the status is ive been shirking my maintainer duties and havent had time to
look at it properly :p

one thing hat can help maoe this happen is pinging the issue or pinging me
on gitter so i dont forget

the other thing that can help is helping me fix the build so it's not red
-.- the  i'll be able to work on stuff with better certainty

acjay:
It would also be great to be able to set this per-class.

hepin1989:
Any update for this @stephanh ?

lihaoyi:
Bug bankruptcy

PR #191, opened 2016-12-28
Added better integration of Long, BigInt, and BigDecimal support

darkfrog26:
In the current version of the code Long, BigInt, and BigDecimal are only supported from Strings. This patch modifies them to support either String or Number in the Reader and updates Long to write to a Number while BigInt and BigDecimal still write to a String.


darkfrog26:
Something is wrong with the Travis configuration...the tests are running fine.

lihaoyi:
Bug bankruptcy

PR #198, opened 2017-01-27
Verify that primary constructors are public and return correct type

shadaj:
When trying to unpickle Squants quantities with uPickle, uPickle tries to use the apply methods of the [companion object](https://github.com/typelevel/squants/blob/master/shared/src/main/scala/squants/motion/Velocity.scala#L49) of the quantity type. With Squants, one of these is private and the other does not return the quantity type (it returns a `Try` of it). This PR prevents uPickle from generating an uncompilable `Reader` for such cases by verifying that the apply method is public and the method returns the appropriate type.


lihaoyi:
Bug bankruptcy

PR #201, opened 2017-02-27
Fix #190

### LINKED ISSUE #190: Unable to read case classes with vararg
``` scala

final case class V(s: String*)

object V {
  import upickle.default._

  read[V](???)

}
```

```
V.scala:8: type mismatch;
 found   : Seq[String]
 required: String
  read[V](???)
         ^
```

tkroman:
Single-argument vararg constructors require a special treatment in a macro for `case1`-Reader (`: _*`).


tkroman:
Build failure seems unrelated to the change since tests pass - looks like a travis issue.

PR #210, opened 2017-05-24
change default implicit Reader[Map] to use ListMap (keeps order)

scalway:
Default implementation of Reader[Map[K,V]] don't keep order of fields in objects. This is information that  we can keep without trouble (probably). I used similar implementation in my code (mostly scripts).

i used this code to validate this pull request:
```scala
import upickle.default._
import scala.collection.immutable.ListMap
val rest:Map[String, String] = (1 to 100).map {
  r => ("r:" + r.toString, "l:" + r.toString)
}(collection.breakOut):ListMap[String, String]

read[Map[String,String]](write(rest))
```


lihaoyi:
Bug bankruptcy

Issue #218, opened 2018-01-30
Why no Js.Bool


SrTobi:
Hi, thanks for the library. It's exactly what I was searching for.
One question, though: Why is there no `Js.Bool`?
Just a simple base class for `Js.True` and `Js.False`. It doesn't even need an implementation. Just apply and unapply...
Example (not tested or anything):
```scala
sealed abstract class Bool extends Value {
  def value: Boolean
}
object Bool {
  def apply(value: Boolean): Bool = if (value) True else False
  def unapply(bool: Bool): Option[Boolean] = bool.value
}
case object False extends Value{
  def value = false
}
case object True extends Value{
  def value = true
}
```
Would have made some things more convenient for me 😄

lihaoyi:
Feel free to send a PR then!

PR #220, opened 2018-03-26
created Js.Bool

SrTobi:
As discussed in #218 I added a Js.Bool as base class for Js.True and Js.False
- added two implicit readers
- added simple api test spec (just to check if it works at all)
- added simple read/write test case


lihaoyi:
lgtm

PR #223, opened 2018-04-02
Added bool helper method for easy boolean access

lbialy:
As in topic and in response to twitter chat :smile:


SrTobi:
:+1: thx, forgot that

PR #227, opened 2018-04-22
Reformat the code of ujson.

ShutaoTang:
I notice that some codes in `ujson` project owns the following style:
```
trait AstTransformer[I] extends Transformer[I] with ujson.Visitor[I, I]{ 
// there is no whitespace between "Visitor[I, I]" and "{"
```
So I add a whitespace before `{`

```
trait AstTransformer[I] extends Transformer[I] with ujson.Visitor[I, I] {
```


lihaoyi:
Probably not worth doing as a one off

PR #238, opened 2018-07-18
WIP port to scala-native

lihaoyi:
Blocked by https://github.com/scala-native/scala-native/issues/1280


LeeTibbert:
@lihaoyi 
Not to be a pest, but did you get notice of my post to your Scala-Native issue #1280?
Just checking for "lost tokens" of communication.

Short story was that I looked at the issue, trying to unblock this port. I could not replicate
using 0.3.7 in an sbt project, so I was looking for help replicating.

I admire your work and would like to help your porting if I can.

Cheers,

Lee

lihaoyi:
@LeeTibbert perhaps you could try replicating it in the existing Mill project, minimizing it there, and then seeing what the difference is between how Mill and SBT use scala-native?

PR #241, opened 2018-08-21
Add scaladoc: Visitor

htmldoug:
## Scaladoc
http://www.lihaoyi.com/post/ZeroOverheadTreeProcessingwiththeVisitorPattern.html inspired me to explore ujson's `AsyncParser` for streaming json transformation (outstanding read, thank you!).

For a newcomer, some bits of the API aren't immediately obvious and would benefit from scaladoc. For example:
```scala
// Is this an index within the CharSequence? (no)
def visitString(s: CharSequence, index: Int): J

// what are all these indexes?!
def visitNum(s: CharSequence, decIndex: Int, expIndex: Int, index: Int): J

// what does "value" refer to?
def visitValue(v: J, index: Int): Unit
```

### DIFF
--- a/ujson/src/ujson/Visitor.scala
+++ b/ujson/src/ujson/Visitor.scala
@@ -1,20 +1,55 @@
 package ujson
 
 /**
- * Facade is a type class that describes how Jawn should construct
- * JSON AST elements of type J.
- *
- * Facade[J] also uses FContext[J] instances, so implementors will
- * usually want to define both.
- */
+  * Visitor is a type class that describes how ujson should visit json elements.
+  *
+  * Visitors return their results type as [[J]] which may be an AST, e.g. [[AstTransformer]],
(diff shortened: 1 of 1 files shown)


ujson/src/ujson/Visitor.scala
```diff
@@ -1,20 +1,55 @@
...
- * Facade[J] also uses FContext[J] instances, so implementors will
- * usually want to define both.
- */
+  * Visitor is a type class that describes how ujson should visit json elements.
+  *
+  * Visitors return their results type as [[J]] which may be an AST, e.g. [[AstTransformer]],
+  * or a writer-like type that Visitors apply side-effects to, e.g. [[StringRenderer]], [[BytesRenderer]].
+  *
+  * @see [[http://www.lihaoyi.com/post/ZeroOverheadTreeProcessingwiththeVisitorPattern.html]]
+  * @tparam T ???
```
htmldoug: I still have no idea about this one.

PR #249, opened 2018-10-09
Fix dead code warning on write macro

### LINKED ISSUE #236: getting dead code following this construct warning while using macroRW
```
case class TcpConnection(name: String, host: String, port: String)
object TcpConnection {
      implicit val tcpConnectionRw: RW[TcpConnection] = macroRW
}
```
This code generate following warnings:
```
dead code following this construct
[warn]   implicit val tcpConnectionRw: RW[TcpConnection]   = macroRW
```

mrdziuban:
Fixes #236.

I'm not fully sure why the macro was previously casting the writer and the visitor using `asInstanceOf`, or why the dead code warnings were happening, but everything seems to be working as expected with this change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 sudo: false
 language: scala
 scala:
-  - 2.11.11
-  - 2.12.4
+  - 2.11.12
+  - 2.12.6
 
 script:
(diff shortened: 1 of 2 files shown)


lihaoyi:
merged in https://github.com/lihaoyi/upickle/commit/f3dbc80be8b1c556e806b729f1b3244d699ac289

PR #264, opened 2019-03-01
Scala Native support

lolgab:
- Removed the dependency from Scalacheck and substituted with Scalatest-3.1.0-SNAP7 that implements the same Generators and converted the tests to scalatest syntax.
- Modified the build to have mixed platforms source folders (for example `src-jvm-js` for source to compile on both jvm and js but not on native.
- Moved some test files to jvm-js only because of different toString behavior of Scala Native 0.3.8 ( for example `"2.0" != "2.0000" ). They should be moved in the shared source folder when this bug goes away.
- Created two empty test suited because of mill bug complaining when there are no tests on native.


lolgab:
I don't understand why travis is failing. Who sets `TRAVIS_SCALA_VERSION` environment variable?

lihaoyi:
Closing due to inactivity, feel free to bring it up to master and I'll re-open it for review

PR #280, opened 2019-08-19
Fix indentation in Api.scala snippet

DieBauer:
The Common Operations section (http://www.lihaoyi.com/upickle/#CommonOperations) has a snippet of `Api.scala`, this snippet misses the first 4 characters of each line, which results in weird highlighting. 

By dropping the `""` in the Seq, the complete file gets rendered correctly.


DieBauer:
Result checked with `sbt upickleReadme/run`

PR #286, opened 2019-09-29
Add JsonPointerVisitor/JsonPointerException

htmldoug:
Port of https://github.com/rallyhealth/upickle/pull/8 to add [JSON Pointer](https://tools.ietf.org/html/rfc6901)s (e.g. `/hits/hits/3/_source`) to exceptions thrown by visitors.

Biggest caveat is that MessagePack reads are [10% slower on java 8](https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/htmldoug/2f99a27a22cc23aab5df751af308daa5/raw/c5595919d44c910e7b4a3ef97a4c0fb815eea416/1.json,https://gist.githubusercontent.com/htmldoug/2f99a27a22cc23aab5df751af308daa5/raw/c5595919d44c910e7b4a3ef97a4c0fb815eea416/2.json) and [20% slower on graalvm-19.2.0](https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/htmldoug/73e673b6b2918014a06ad1d085fb56b6/raw/e59788a83b5012e9ee9e294aadcbbad1efabe2ea/1-graalvm-19.json,https://gist.githubusercontent.com/htmldoug/73e673b6b2918014a06ad1d085fb56b6/raw/e59788a83b5012e9ee9e294aadcbbad1efabe2ea/2-graalvm-19.json). JFR shows that 20% of the allocations now are `JsonPointerVisitor` instances.

It's strange that adding `JsonPointerVisitor` consistently makes regular string json reads faster. I have no explanation for that.

I'm open to ideas for the MessagePack perf. We could just remove from MessagePack where people are more likely to be concerned about perf. They can add it themselves if desired, although the same could be said for removing it.


lihaoyi-databricks:
Not terribly surprised that adding another visitor slows things down. Every layer of indirection costs something, and the baseline parsers are reasonably tightly optimized.

How about make usage of `JsonPointerVisitor` opt-in, so people can decide for themselves if it's worth it? e.g. leaving the default `upack.transform`/`ujson.transform`/etc. bare, but if someone wants better error messages they can always use `transform(JsonPointerVisitor(...))` themselves to pay the cost and get pretty errors

lihaoyi-databricks:
Another question: this only annotates exceptions in the downstream visitor, and not the upstream source of JSON events, is that right? e.g. that means that a parse error halfway through a document won't get annotated with metadata. That's probably fine if that's the case, but we should be aware of that and maybe put it in the doc-comment

core/src/upickle/core/JsonPointerVisitor.scala
```diff
@@ -0,0 +1,189 @@
...
+    def path: String = components.iterator.map(_.pathComponent).flatten.mkString("/")
+
+    override def toString: String = path
+
+    def wrap[T](f: => T): T = {
+      try {
+        f
+      } catch {
+        case NonFatal(cause) =>
+          cause.addSuppressed(new JsonPointerException(path))
```
htmldoug: This would be prettier unsuppressed, but I didn't want to update all the tests or break anyone out there that's catching an `Abort`.

htmldoug:
> this only annotates exceptions in the downstream visitor, and not the upstream source of JSON events

That's right. I've slightly tweaked the scaladoc.

htmldoug:
> How about make usage of `JsonPointerVisitor` opt-in

Also done.

lihaoyi:
Last question: should we go with a `JsonPointerException` that wraps the inner exception via `cause`, rather than doing this `addSuppressed` thing? My impression is that `addSuppressed` is a pretty esoteric feature that most people will not be familiar with, whereas everyone knows what wrapped exceptions are and what they mean

htmldoug:
Done. That was my preference, too. I'm surprised that changing the outermost exception didn't break any tests.

htmldoug:
Actually, hold on. Turning it on by default exposes a coupling in `Parser` that expects an `AbortException`: https://github.com/lihaoyi/upickle/blob/master/ujson/src/ujson/Parser.scala#L362

This is what I feared when I initially made it suppressed.

htmldoug:
Why does `Abort` get special treatment here? Shouldn't any visitor-thrown `Throwable`s get index information?

lihaoyi:
@htmldoug good question. The only thing special about `Abort` is that it's meant to be a "known" exception and thus not need a stack trace, v.s. an unexpected exception that would come with a stack trace. I suppose it would be fine to remove that special casing of `Abort`, make unexpected exceptions get annotated with index information as well

htmldoug:
Removing the special casing of `Abort` in `Parser` makes other tests fail. At the moment I'm out of time to work through those.

I've merged https://github.com/rallyhealth/upickle/pull/8 with the `JsonPointerException` suppressed for now, but left open the `cause: Throwable` param to set later if the `Parser` assumptions are removed.

Would you like for me to make the same changes in this PR?

lihaoyi-databricks:
@htmldoug ok, I'll take a look at removing the special casing for `Abort`. I assume it would just be a matter of updating the test cases to handle the wrapped exceptions being thrown

htmldoug:
Closed due to age. Holler if you want me to reopen.

lihaoyi:
@htmldoug I ended up merging this with some modifications: 

- JsonPath-style syntax rather than JsonPointer
- Using mutable variables rather than sprinkling try-catch blocks throughout the visitor
- Opt in for `ujson.read`, `upickle.default.read`, `upack.read` using the `trace = true` flag
- Support for providing traces for parse/syntax errors as well

```scala
@ ujson.read("""{"hello":""", trace=true)
upickle.core.TraceVisitor$TraceException: $['hello']

ujson.IncompleteParseException: exhausted input
  ujson.Parser$$anonfun$1.applyOrElse(Parser.scala:371)
  ujson.Parser$$anonfun$1.applyOrElse(Parser.scala:369)
  scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:35)
  ujson.Parser$$anonfun$reject$1.applyOrElse(Parser.scala:374)
  ujson.Parser$$anonfun$reject$1.applyOrElse(Parser.scala:374)
  scala.PartialFunction$OrElse.apply(PartialFunction.scala:238)
  ujson.Parser.parse(Parser.scala:369)
  ujson.Parser.parse(Parser.scala:104)
  ujson.StringParser$.transform(StringParser.scala:29)
  ujson.StringParser$.transform(StringParser.scala:28)
  ujson.Readable$fromTransformer.transform(Readable.scala:13)
  ujson.package$.transform(package.scala:4)
  ujson.package$.$anonfun$read$1(package.scala:10)
  upickle.core.TraceVisitor$.withTrace(TraceVisitor.scala:23)
  ujson.package$.read(package.scala:10)
  ammonite.$sess.cmd4$.<clinit>(cmd4.sc:1)

@ upickle.default.read[Seq[Foo]]("""[{"x": 1, "s": ["a"]}, {"x": 2, "s": [1]}]""", trace=true)
upickle.core.TraceVisitor$TraceException: $[1]['s'][0]

upickle.core.AbortException: expected string got number at index 38
  ujson.Parser$$anonfun$reject$1.applyOrElse(Parser.scala:378)
  ujson.Parser$$anonfun$reject$1.applyOrElse(Parser.scala:374)
  scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:35)
  ujson.Parser.rparse(Parser.scala:423)
  ujson.Parser.parse(Parser.scala:346)
  ujson.Parser.parse(Parser.scala:104)
  ujson.StringParser$.transform(StringParser.scala:29)
  ujson.StringParser$.transform(StringParser.scala:28)
  ujson.Readable$fromTransformer.transform(Readable.scala:13)
  upickle.Api.$anonfun$read$1(Api.scala:40)
  upickle.core.TraceVisitor$.withTrace(TraceVisitor.scala:23)
  upickle.Api.read(Api.scala:40)
  upickle.Api.read$(Api.scala:39)
  upickle.default$.read(Api.scala:131)
  ammonite.$sess.cmd3$.<clinit>(cmd3.sc:1)
upickle.core.Abort: expected string got number
  upickle.core.SimpleVisitor.visitFloat64StringParts(SimpleVisitor.scala:17)
  upickle.core.SimpleVisitor.visitFloat64StringParts$(SimpleVisitor.scala:16)
  upickle.implicits.Readers$$anon$9.visitFloat64StringParts(Readers.scala:90)
  upickle.core.Visitor$Delegate.visitFloat64StringParts(Visitor.scala:158)
  ujson.Parser.parseNum(Parser.scala:188)
  ujson.Parser.rparse(Parser.scala:423)
  ujson.Parser.parse(Parser.scala:346)
  ujson.Parser.parse(Parser.scala:104)
  ujson.StringParser$.transform(StringParser.scala:29)
  ujson.StringParser$.transform(StringParser.scala:28)
  ujson.Readable$fromTransformer.transform(Readable.scala:13)
  upickle.Api.$anonfun$read$1(Api.scala:40)
  upickle.core.TraceVisitor$.withTrace(TraceVisitor.scala:23)
  upickle.Api.read(Api.scala:40)
  upickle.Api.read$(Api.scala:39)
  upickle.default$.read(Api.scala:131)
  ammonite.$sess.cmd3$.<clinit>(cmd3.sc:1)

```

htmldoug:
Thanks for letting me know. Happy to give back in some way! I'll review your implementation for goodies.

weePickle currently gets these paths from jackson-core, but I've been experimenting with jsoniter-scala in performance-sensitive places. The speedups are significant, but I'd still need to add this Visitor to get error paths.

htmldoug:
https://github.com/rallyhealth/weePickle/pull/13 for the jsoniter-scala integration and benchmarks, if interested.

lihaoyi:
Yeah I'd love your review on my implementation. Hopefully I covered all the cases, but these visitor-style APIs are tricky and it's easy to miss a spot. I tried to flesh out the test suite to cover the various edge cases, but more eyes are always welcome.

I'd love to get a faster parser. The uPickle builtin parser does ok, but it's a bit of a janky piece of code I copied for Jawn half a decade ago with neither understanding nor appreciation. I have no doubt that it can be replaced by something better, and with uPickle's architecture we should be able to swap in a new parser with ~zero change to the rest of the library (similar to how you're using Jackson)

htmldoug:
I have a Visitor TCK mostly written in a branch somewhere. I'll see about tidying it up and sharing.

I'm curious to see if you'd inline jsoniter-scala or take the dependency on an external jar. Unlike jackson it is published for scala JS. There are some crazy fast but complicated optimizations in there that I'd be scared to maintain.

lihaoyi:
I'd prefer to inline the necessary parts. It's a bit scary to maintain, but can't be that much more scary than the weird Jawn parser that we had before with it's buggy async parsing and stuff.

We can also try vendoring it while preserving the code mostly intact, so in future we can simply copy-paste the latest parser from Jawn or whatever into our codebase without needing to get into the details of the code ourselves.

There's also a Borer parser from https://github.com/sirthias/borer, which is supposed to be quite fast. Not sure how it compares with Jsoniter

htmldoug:
The only issue I could find with `TraceVisitor` was one path reported differently with the changes to `visitArray` index. Test fails [here](https://github.com/rallyhealth/weePickle/blob/458775e3d64f6234d721d250120c177d077c6acd/weepickle/test/src/com/rallyhealth/weepickle/v1/JsonPointerVisitorTests.scala#L76):

```
assertion failed: ==> assertion failed: /foo/0 != /foo
java.lang.AssertionError: assertion failed: ==> assertion failed: /foo/0 != /foo
	at scala.Predef$.assert(Predef.scala:223)
	at utest.asserts.Asserts$ArrowAssert.$eq$eq$greater(Asserts.scala:195)
	at com.rallyhealth.weepickle.v1.JsonPointerVisitorTests$.$anonfun$tests$17(JsonPointerVisitorTests.scala:76)
```

In the scenario where `underlying.visitArray()` throws, the path would point to the first array element rather than with the array (i.e. `[`) itself. That's what the [i => 0 check](https://github.com/rallyhealth/weePickle/blob/458775e3d64f6234d721d250120c177d077c6acd/core/src/com/rallyhealth/weepickle/v1/core/JsonPointerVisitor.scala#L159) is meant to prevent.

Otherwise, `TraceVisitor` LGTM.

My Visitor tests are mostly concerned that the underlying methods are invoked in the right order. There's nothing interesting happing with control flow here, so I'm going to leave those on the shelf for now.

Did you see any perf changes with the single try block and IoC or was that just to simplify? 20% penalty seems about the same as I was seeing.

PR #291, opened 2020-01-03
Add option for serializing default values also

sake92:
This adds an annotation so that default values can be serialized by default.  
I chose to call it `serialize` if we need another flag in the future, similar to what Jackson does e.g. ...

### DIFF
--- a/implicits/src/upickle/implicits/internal/Macros.scala
+++ b/implicits/src/upickle/implicits/internal/Macros.scala
@@ -5,6 +5,7 @@ import scala.language.experimental.macros
 import compat._
 import acyclic.file
 import upickle.implicits.key
+import upickle.implicits.serialize
 
 import language.higherKinds
 import language.existentials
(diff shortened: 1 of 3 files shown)


lihaoyi:
Thanks @sake92! Do you think we could make it a configuration option on `upickle.Api`, rather than on each individual field? I assume most people who want to configure this would want it to be configured consistently across all their types, rather than configured on a field-by-field basis

sake92:
@lihaoyi 
IDK how that approach would look like, how would we propagate the information about the default values?

---
An "aside", this feature is unintuitive to me. I haven't seen it in any other library.  
If you have a default value and create an object, the value is there, serialization should just serialize it... It's very confusing to everyone, there's one special value that is not serializable, and that's the default value... 😕   
Example: https://scalafiddle.io/sf/u1DnNTT/2 (I can never serialize the value `"display":"abc"`)


Maybe instead we should use something like Jackson's `@JsonIgnore` annotation to avoid serializing/deserializing the field? They use getters for ignoring serialization and setters for deserialization. We could use something like `@serialization(ignore=true)` and  `@deserialization(ignore=true)` (although we should avoid `null`s)?

Here the approach would be flipped. Always (de)serialize unless explicitly said so.  
That's more natural to me.

lihaoyi:
@sake92 Confusing or not, It's been like this for 5 years now, so I'm not willing to change it and cause all existing users trying to upgrade silent breakage. The time to debate this kind of thing was in early 2015 😛 

To move it to the `Api` trait, we simply define a method there `serializeDefaults: Boolean`, refer to it in the macro via `${c.prefix}`, and perform a runtime check whether or not we should serialize the default values at the part where we're comparing the value with the default for equality.

sake92:
Haha, fair enough! 😄  
Closing in favor of #295

PR #293, opened 2020-01-08
Perf: Avoid unnecessary buffer reallocation

htmldoug:
Port of https://github.com/rallyhealth/weePickle/pull/36.

There's far more `Array[Byte]` allocation happening in `BufferingInputStreamParser` than there should be. 1.56 GB total / 25,508 objects = 64 KB avg size.

Screen Shot 2020-01-08 at 5 16 41 PM

Debugger reveals that when `buffer.length` is 65,536, and `newSize` is 65,536, implementation reallocates `buffer` every time just to left shift the contents. Avoidable.

BufferingInputStreamParser requestUntil

### DIFF
--- a/core/src/upickle/core/BufferingInputStreamParser.scala
+++ b/core/src/upickle/core/BufferingInputStreamParser.scala
@@ -69,7 +69,7 @@ trait BufferingInputStreamParser{
       val growGoalSize = (until - dropped + 1)
       while (newSize <= growGoalSize) newSize *= 2
 
-      val arr = if (newSize > buffer.length / 2) new Array[Byte](newSize) else buffer
+      val arr = if (newSize > buffer.length) new Array[Byte](newSize) else buffer
 
       System.arraycopy(buffer, dropped - firstIdx, arr, 0, lastIdx - dropped)
       firstIdx = dropped
--- a/core/src/upickle/core/BufferingInputStreamParser.scala
+++ b/core/src/upickle/core/BufferingInputStreamParser.scala
@@ -63,7 +63,7 @@ trait BufferingInputStreamParser{
     //
     // - https://github.com/scala-js/scala-js/issues/3913
     val untilBufferOffset = until - firstIdx + 1
-    if (untilBufferOffset >= buffer.size){
+    if (untilBufferOffset >= buffer.length){
       var newSize = buffer.length
 
       val growGoalSize = (until - dropped + 1)


htmldoug:
Also, thank you so much for optimizing the `InputStream`-based stuff. The buffer size of that old `ChannelParser` was enormous, which made it unsuitable for small msgs. The new 64 KB size is fantastic.

htmldoug:
Hold on, I don't think this zeros the array.

htmldoug:
I guess it's fine. The buffer contents look silly after left-shifting, but as long as the indicies are correct, there's no problem. If the indicies are wrong, we've got bigger problems anyway.

lihaoyi:
we should not need to zero the arrays while doing streaming parsing; it
already keeps track of the valid length of bytes IIRC

core/src/upickle/core/BufferingInputStreamParser.scala
```diff
@@ -63,13 +63,13 @@ trait BufferingInputStreamParser{
     //
     // - https://github.com/scala-js/scala-js/issues/3913
     val untilBufferOffset = until - firstIdx + 1
-    if (untilBufferOffset >= buffer.size){
+    if (untilBufferOffset >= buffer.length){
```
htmldoug: Seems like the jvm should inline this away, but it's still showing up under profiler and easy to fix.

buffer size

lihaoyi:
@htmldoug this `/ 2` is intentional; see the comment above. The goal of that was to try and avoid a degenerate case where every time we free a small number of bytes via `dropBufferUntil` and ask for another small number of bytes via `requestUntil`, we end up left-shifting an arbitrarily large number of bytes in the buffer. 

In the worst case, every byte being parsed could require copying `O(buffer.length)` bytes to perform the left shifting. Hence this `/ 2` factor was meant to ensure that the buffer is never more than half full, so in the worst case the `num_bytes_copied/num_bytes_parsed` can only max out at `2`.

In the end this should result in a buffer that ranges from 2-4 times the necessary parsing window, but my understanding is that JSON parsing windows tend to be pretty small unless you have huge string literals. In particular, a 64kb JSON blob should not require a 64kb buffer because of frequent calls to `dropBufferUntil`. What kind of scenarios are you hitting that the oversize buffer becomes a problem?

htmldoug:
Thanks for explaining the motivation. After rebasing, something still seems fishy. I'm testing against `new ByteArrayInputStream(Common.benchmarkSampleJson.getBytes()))`, which is 671,001 long.

The following screenshot captures the first reallocation, which I don't think should happen, since the buffer would be at < 0.1% capacity after an arraycopy in place (lastIdx=65535 - dropped=65504 = 32 unread bytes). Wouldn't it be cheaper to copy 32 bytes to the start of the array than to allocate a whole new array (of the exact same size) and copy the 32 bytes into that?

Screen Shot 2020-01-12 at 7 28 18 PM

This got me second guessing whether arraycopy() is really able to copy in place from the same src and dest. The System.arraycopy() javadocs say:

> If the src and dest arguments refer to the same array object, then the copying is performed *as if* the components at positions srcPos through srcPos+length-1 were first copied to a temporary array with length components and then the contents of the temporary array were copied into positions destPos through destPos+length-1 of the destination array.

I wrote up an isolated benchmark to prove to myself that "as if" doesn't actually mean it it creates the temporary array:
```
private val size: Int = 64 * 1024
val arr = Array.ofDim[Byte](size)
Random.nextBytes(arr)

@Benchmark
def arrInPlace(bh: Blackhole): Unit = {
  val destPos = Random.nextInt(1000)
  System.arraycopy(arr, 65504, arr, destPos, 32)
  bh.consume(arr)
}

@Benchmark
def arrAlloc(bh: Blackhole): Unit = {
  val reallocated = Array.ofDim[Byte](size)
  System.arraycopy(arr, 65504, reallocated, 0, 32)
  bh.consume(reallocated)
}
```

```
Benchmark             Mode  Cnt         Score         Error  Units
JmhBench.arrAlloc    thrpt   10    254182.508 ±   18303.748  ops/s
JmhBench.arrInPlace  thrpt   10  78194162.071 ± 1620220.037  ops/s
```

Admittedly, once all the other json processing is added back in, the difference is harder to quantify. Even so, it seems possible to avoid both the degenerate case, and the array reallocation at 0.1% capacity.

lihaoyi:
Yeah, I wouldn't be surprised i there was some misbehavior somewhere in the streaming buffering logic. If you can figure out exactly what is going wrong we could fix it while preserving the existing intended behavior

htmldoug:
Closing this for now.

PR #295, opened 2020-01-12
Serialize defaults flag in Api

sake92:
This is a replacement for #291  
@lihaoyi I think this is what you had in mind?


lihaoyi:
@sake92 yeah this looks good, thanks!

PR #297, opened 2020-01-17
Correctness over mutable java.lang.CharSequence inputs.

htmldoug:
I found a corner-case correctness issue when feeding [mutable](https://github.com/rallyhealth/weePickle/blob/b088175f380bd47e830376b909e064052c85cdfc/weejson/jackson/src/com/rallyhealth/weejson/v0/jackson/VisitorJsonGenerator.scala#L66-L71) `CharSequence`s into the generated macros.

Input of `"""{"s1":"a","s2":"b", "$type": "upickle.Hierarchy.C"}"""` fails in this scenario. Macro ends up with values of `Str("upickle.Hierarchy.C")` for "s1" and "s2".

Freezing `Str` to `CharSequence.toString` resolves the issue. We'd also need `StringVisitor` to freeze to a `String` since it's used by the macro for parsing the value of `"$type"`.

AFAIK, the ujson parsers only emit `java.lang.String` anyway, so I expect the cost should be negligible. As long as that's true, it's only an issue for extensibility outside the library.

### DIFF
--- a/core/src/upickle/core/StringVisitor.scala
+++ b/core/src/upickle/core/StringVisitor.scala
@@ -2,5 +2,5 @@ package upickle.core
 
 object StringVisitor extends SimpleVisitor[Nothing, Any] {
   def expectedMsg = "expected string"
-  override def visitString(s: CharSequence, index: Int) = s
+  override def visitString(s: CharSequence, index: Int): String = s.toString
 }
--- a/ujson/src/ujson/IndexedValue.scala
+++ b/ujson/src/ujson/IndexedValue.scala
@@ -88,6 +88,6 @@ object IndexedValue extends Transformer[IndexedValue]{
     def visitFloat64StringParts(s: CharSequence, decIndex: Int, expIndex: Int, i: Int) = IndexedValue.Num(i, s, decIndex, expIndex)
     override def visitFloat64(d: Double, i: Int) = IndexedValue.NumRaw(i, d)
 
-    def visitString(s: CharSequence, i: Int) = IndexedValue.Str(i, s)
+    def visitString(s: CharSequence, i: Int) = IndexedValue.Str(i, s.toString)
   }
 }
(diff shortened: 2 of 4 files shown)


lihaoyi:
I don't quite understand this PR. Isn't the point of accepting CharSequence that we can take in non-string inputs without conversions? I feel like passing in mutable `CharSequences` is OK, on a "buyer beware" basis, as long as you don't mutate it between being passed in and being used: similar to putting mutable data structures into a set or as dictionary keys.

If you want to make a defensive copy of a mutable `CharSequence`, you can do that before passing it to uPickle, without needing uPickle itself to make such copies. Similar to how `Set` and `Map` do not make defensive copies of their keys/entries

htmldoug:
>  Isn't the point of accepting CharSequence that we can take in non-string inputs without conversions? 

Yes, exactly.

> as long as you don't mutate it between being passed in and being used

The problem is that the parser has no idea which `CharSequence`s uPickle requires to be immutable and which can be mutable. When the `"$type"` field is not the first to arrive in an object, uPickle holds references to the mutable `CharSequences` of both keys and values until later within `IndexedValue`, but expects the contents not to change, and does not make defensive copies itself.

If it's the parser's responsibility, how can it determine in which situations uPickle requires defensive copies, particularly before it even parses a `"$type"` field?

htmldoug:
Running the new test case without the `StringVisitor` and `IndexedValue` changes might make it clearer why it's currently unsafe to pass mutable `CharSequence` inputs to uPickle at all. We need them frozen by the macros whenever they buffer input into `IndexedValue`. Otherwise, the contents of those `CharSequence`s change by the time the macros replay them into the real `Visitor`.

uPickle is doing the buffering in this case, so it's up to uPickle to make defensive copies with `.toString`.

htmldoug:
Closing due to age.

PR #298, opened 2020-01-18
Bugfix: EitherWriter writes outside of array.

htmldoug:
I have a Visitor that prevents further writes after a single value (e.g. okay: `["1", 2]`, but not okay: `"1" [2]`). In this case, writing the "1" to the root visitor instead of the array visitor prematurely triggers a close() on the underlying resource. The encoding for `Left(2)` is `["1" , 2]` regardless, since the `ctx.visitValue()` call is correct, but the EitherWriter was sending the `.visitFloat64StringParts("1")` to the wrong place which causes problems if visiting the root prevents multiple calls.

Found during https://github.com/rallyhealth/weePickle/pull/52.


lihaoyi:
looks good to me, thanks!

Issue #303, opened 2020-02-17
Potential file handle leak


lefou:
We found this downstream in mill and track it as https://github.com/lihaoyi/mill/issues/754 .

@ajrnz was able to track it down to upickle, somewhere between 0.9.0 and 0.9.5. I do just quote his comment (https://github.com/lihaoyi/mill/issues/754#issuecomment-586977490) here:
> I've just had a quick look at this. It appears to be a problem somewhere under `upickle.default.read`.
> 
> Here is a minimized test case:
> 
> ```scala
> object MillTest extends App {
>   //val baseDir = os.pwd
>   val baseDir = os.home / "dev" / "ext" / "mill"
>   val x =
>     os.walk(baseDir)
>       .filter(_.last == "meta.json")
>       .map{ p =>
>       println(s"Loading: $p")
>       upickle.default.read[ujson.Value](p.toIO)
>     }
> 
>   println("waiting")
>   Thread.sleep(1000*1000)
> }
> ```
> 
> Using:
> 
> ```
>     ivy"com.lihaoyi::upickle:0.9.5",
>     ivy"com.lihaoyi::os-lib:0.6.2"
> ```
> 
> If you find the java PID and then do a `lsof -p <pid>` you see all the files are open.
> 
> Appears to have been introduced between upickle 0.9.0 and 0.9.5.
> 
> Don't have more time to dig at the moment but hope this helps...

nafg:
Poking around the code, this seems to come down to https://github.com/lihaoyi/upickle/blob/master/ujson/src/ujson/Readable.scala#L22 which uses https://github.com/lihaoyi/upickle/blob/master/ujson/src/ujson/InputStreamParser.scala which has a no-op `close` method. Is there a reason it doesn't call `data.close()` instead?

I notice that the Transformer instance calls `parse` which calls `close` at https://github.com/lihaoyi/upickle/blob/master/ujson/src/ujson/Parser.scala#L114.

Perhaps some `InputStreamParser`s need to not close the InputStream and some need to close it?

lihaoyi:
@nafg you are right. I think the logic to close the input stream should live in https://github.com/lihaoyi/upickle/blob/master/ujson/src/ujson/Readable.scala, since it's conceivable that an `InputStreamParser` may be used on a longer inputstream, whereas a `Readable` is defined to be a one-shot affair that cleans up after itself.

nafg:
[...]

PR #304, opened 2020-02-18
ujson.Readable.from{Path,File}: close input stream

nafg:
Attempt at fixing #303 

What's confusing about the code (both before and after) is that it creates an InputStream once, and then returns an object that can attempt to consume the InputStream many times. For this reason closing it feels wrong but in any case the InputStream will presumably be useless the next time.

(Unless it's meant to be usable with an `ndjson` file. In which case #303 has no solution.)

I would think a better approach would be that every time `Readable#transform` is called it would create the InputStream from the file, read it, and close the InputStream. But that doesn't seem to fit the whole InputStreamParser + Transformer shape of things.

### DIFF
--- a/ujson/src/ujson/Readable.scala
+++ b/ujson/src/ujson/Readable.scala
@@ -15,12 +15,11 @@ object Readable extends ReadableLowPri{
   }
   implicit def fromString(s: String) = new fromTransformer(s, StringParser)
   implicit def fromCharSequence(s: CharSequence) = new fromTransformer(s, CharSequenceParser)
-  implicit def fromPath(s: java.nio.file.Path) = {
-    val inputStream = java.nio.file.Files.newInputStream(s)
-    new fromTransformer(inputStream, InputStreamParser) {
-      override def transform[T](f: Visitor[_, T]) =
-        try super.transform(f)
-        finally inputStream.close()
+  implicit def fromPath(s: java.nio.file.Path) = new Readable {
+    override def transform[T](f: Visitor[_, T]) = {
+      val inputStream = java.nio.file.Files.newInputStream(s)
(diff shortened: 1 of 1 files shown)


lihaoyi:
Thanks for the PR, it's a bit late here now but I'll take a look at it soon

lefou:
I just wanted to let you know that this PR seem to fix the issue in mill (https://github.com/lihaoyi/mill/issues/754). At least, my test script does not complain.

ujson/src/ujson/Readable.scala
```diff
@@ -15,14 +15,15 @@ object Readable extends ReadableLowPri{
...
-    InputStreamParser
-  )
-  implicit def fromFile(s: java.io.File) = new fromTransformer(
-    java.nio.file.Files.newInputStream(s.toPath),
-    InputStreamParser
-  )
+  implicit def fromPath(s: java.nio.file.Path) = {
+    val inputStream = java.nio.file.Files.newInputStream(s)
+    new fromTransformer(inputStream, InputStreamParser) {
+      override def transform[T](f: Visitor[_, T]) =
```
lihaoyi: @nafg since we are overriding `def transform`, could we skip `fromTransformer` entirely and instead inherit directly from `Readable`? That would allow us to put the `val inputStream` inside the `def transform` function, making it open a new `java.io.InputStream` every time `transform` is called

nafg: I don't see why not, could you take care of it?

lihaoyi:
Looks good to me, thanks @nafg!

lihaoyi:
@lefou I've published this as uPickle 1.0.0

PR #317, opened 2020-06-16
added 'readEither' and 'readTry'

peter-empen:


lihaoyi:
Probably going to close this since it's easily implementable in user-land, and the clutter from wrapping every single exception-throwing method in Either/Try variants would be problematic

PR #322, opened 2020-08-26
Dotty port

jodersky:
- changes to the build configuration, to cross-build most projects against Dotty (plus all the existing versions).
- reimplementation of the Reader/Writer macros. Note that this required some changes to upickle's class hierarchy.

Note that some tests have not been ported:

- some advanced ADT tests, because implicit lookup has changed slightly under Dotty, and some cases do not behave the same and/or lead to search scope explosion
- the legacy tests

The goal of these changes is to port upickle to Dotty while keeping the behaviour as close as reasonably possible to the original. Once this is merged, we can add more dotty-specific features, such as support for the `derives` keyword.


jodersky:
@lihaoyi, I know that this change is huge. Let me know if you'd like to have a live walkthrough

lihaoyi:
shipit

PR #325, opened 2020-10-03
Add extensions to support Dotty's 'derives' syntax

jodersky:
Unfortunately the new extension method mechanism cannot be used, due to some limitations (as described in the comments).

### DIFF
--- a/implicits/src-0/upickle/implicits/CaseClassReader.scala
+++ b/implicits/src-0/upickle/implicits/CaseClassReader.scala
@@ -79,4 +79,9 @@ trait CaseClassReaderPiece extends MacrosCommon:
 
   inline given [T <: Singleton: Mirror.Of] as Reader[T] = macroR[T]
 
+  // see comment in MacroImplicits as to why Dotty's extension methods aren't used here
+  implicit class ReaderExtension(r: Reader.type):
+    inline def derived[T](using Mirror.Of[T]): Reader[T] = macroR[T]
+  end ReaderExtension
+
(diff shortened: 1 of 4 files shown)


jodersky:
cc @lihaoyi, @anatoliykmetyuk

anatoliykmetyuk (commented):
Why add the `derived` method via implicits or extension methods? Why not add it directly to the target object?

jodersky:
`Reader`s and `Writer`s are defined in the ujson module, which is does not depend on macros. Short of shuffling project dependencies (and also actually the trait inheritances that make up a upickle API), I don't see how it can be achieved without using extensions

anatoliykmetyuk (approved):

jodersky:
@lihaoyi, I think you're the only one who has permission to merge this. Would it also be possible to tag and release a new patch version with this? Thanks!

lihaoyi:
@jodersky @anatoliykmetyuk I gave you both commit access, feel free to make use of it for all this Dotty-related work.

On a related note, this is the first time I've realized that i'm the only one with commit access to the repo, and I gave @htmldoug access as well since he's made some contributions in the past and maintains the https://github.com/rallyhealth/weePickle fork

jodersky:
Thanks!

jodersky:
@lihaoyi I also tagged these changes under 1.2.2. If you have the time, could you publish these to maven central please?

lihaoyi:
send 1.2.2 to maven central

PR #327, opened 2020-11-25
Bump Scala 3 to 3.0.0-M2

anatoliykmetyuk:


build.sc
```diff
@@ -38,7 +38,7 @@ trait CommonModule extends ScalaModule {
 }
 trait CommonPublishModule extends CommonModule with PublishModule with CrossScalaModule{
   def publishVersion = "1.2.2"
-  def isDotty = crossScalaVersion.startsWith("0")
+  def isDotty = crossScalaVersion.startsWith("0") || crossScalaVersion.startsWith("3")
```
jodersky, line 41: nitpick: should we rename this to `isScala3`?

jodersky (approved):
Thanks for updating this!

build.sc
anatoliykmetyuk: Yep, this should be done at some point.

PR #328, opened 2020-11-29
Upgrade to scalatest 3.2, scalacheck 1.15

smarter:
And drop support for Scala.js 0.6 as it is no longer supported by
scalacheck.

Also upgrade Scala.js to 1.3 since that's what scalacheck requires.


lihaoyi:
Rendered moot since I ripped out scalatest and scalacheck

PR #329, opened 2020-12-01
Do not use deprecated scala.deriving.ArrayProduct anymore.

sjrd:
Instead, wrap the array in a `Product` explicitly.

### DIFF
--- a/implicits/src-3/upickle/implicits/CaseClassReader.scala
+++ b/implicits/src-3/upickle/implicits/CaseClassReader.scala
@@ -1,7 +1,7 @@
 package upickle.implicits
 
 import compiletime.{summonInline}
-import deriving.{ArrayProduct, Mirror}
+import deriving.Mirror
 import upickle.core.{ Visitor, ObjVisitor, Annotator }
 
 trait CaseClassReaderPiece extends MacrosCommon:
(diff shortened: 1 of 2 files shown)


lihaoyi:
I don't know much about the new Scala 3 APIs, @anatoliykmetyuk @jodersky could you review this change?

implicits/src-3/upickle/implicits/CaseClassReader.scala
```diff
@@ -64,7 +64,13 @@ trait CaseClassReaderPiece extends MacrosCommon:
...
             throw new upickle.core.Abort("missing keys in dictionary: " + missingKeys.mkString(", "))
           }
-          m.fromProduct(ArrayProduct(values.toArray))
+
+          val valuesArray = values.toArray
+          m.fromProduct(new Product {
+            def canEqual(that: Any): Boolean = true
+            def productArity: Int = valuesArray.length
+            def productElement(i: Int): Any = valuesArray(i)
+          })
```
anatoliykmetyuk, line 73: It does look like ArrayProduct was useful after all.

anatoliykmetyuk (approved):
LGTM provided that the removal of ArrayProduct from the API has strong arguments in favour.

implicits/src-3/upickle/implicits/CaseClassReader.scala
sjrd: I or we never said it wasn't useful at all. Just that its usefulness was too limited to deserve a place in the standard library, compared to all the other things that are not there and that could be useful too (even just for deriving).

lihaoyi:
@anatoliykmetyuk I don't remember if I gave you publishing credentials, but feel free to tag/publish a new version if I did, and if I didn't give you credentials let me know if you'd like me to tag/publish a new version and I will do so

anatoliykmetyuk:
I already published all the ported projects for M2. Getting this change in will have to wait until M3.

PR #335, opened 2021-02-28
Reduce generated code size of macros by moving generated code into helpers

lihaoyi:
This should give us all the "easy" wins, without significant performance impact. Anything further would probably require converting local variables and fields into dynamic dictionaries, which would slow things down considerably

### DIFF
--- a/core/src/upickle/core/Types.scala
+++ b/core/src/upickle/core/Types.scala
@@ -197,6 +197,7 @@ trait Types{ types =>
           found |= (1L << currentIndex)
         }
       }
+      def visitKey(index: Int) = _root_.upickle.core.StringVisitor
       protected def storeValueIfNotFound(i: Int, v: Any) = {
         if ((found & (1L << i)) == 0) {
           found |= (1L << i)
--- a/core/src/upickle/core/Types.scala
+++ b/core/src/upickle/core/Types.scala
@@ -186,6 +186,7 @@ trait Types{ types =>
 
   abstract class CaseR[V] extends SimpleReader[V]{
     override def expectedMsg = "expected dictionary"
+
     trait CaseObjectContext extends ObjVisitor[Any, V]{
       def storeAggregatedValue(currentIndex: Int, v: Any): Unit
       var found = 0L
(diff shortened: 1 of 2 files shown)


raquo:
@lihaoyi Thanks a lot! I tested this with Scala 2.13.4 and Scala.js 1.4.0, comparing upickle 1.2.3 with the commit in this PR.

Overall this seems like a 10-20% reduction in weight (hard to say precisely) per _affected_ macroRW.

My project's shared module:
- 411.97KB -> 393.72KB (-18.25KB)
- I have ~87 usages of macroRW, so dumb average is approx. 200 bytes savings per macroRW
- BUT, not every macroRW benefitted from this change. Those that did, shaved off approx 300-600 bytes each.
- I also have maybe 10 non-macroRW ReadWriters, made either with methods like `bimap`, but their sizes weren't affected.

Overhead:
- upickle/Types.scala: 13.43KB -> 22.35KB (+8.92KB)
- Scala.js library usage: 167.8KB -> 162.7KB (-4.9KB)

Looks like a good deal to me.

Note: this project of mine is buildable and webpackable, but not _runnable_ at the moment (I got myself into an unrelated deep mess...) so I couldn't test this branch, I only looked at the bundle size difference.

lihaoyi:
@raquo I pushed two more commits; could you try both of them to see how the bundle size is affected? They should be small improvements, but for this sort of thing every bit counts.

Don't worry about correctness; I trust our test suite to catch things if anything is amiss

raquo:
Of course! Thanks for looking into this.

From 1.2.3 to cfa299f – see my previous comment

From cfa299f to 62e2137 – Moved about 1KB of weight from upickle overhead to my shared module, total reduction approx 100 Bytes in total bundle size

From 62e2137 to 4efbe56 – Moved about 20 bytes from Scala.js library overhead into upickle overhead, no measurable difference to total bundle size

So, I'm afraid, no measurable difference with these two, on my project at least.

lihaoyi:
Here's the `fastOpt` and `fullOpt` bundle byte-sizes from the upickle test suite:

| commit | fastopt | fullopt |
|---:|---:|---:|
| `master` | 6,335,072  | 1,070,766 |
| `reduce generated code size of macros by moving generated code into helpers` | 6,335,072  | 1,051,336 |
| `move macro-generated visitKey into library code` | 6,329,138 | 1,051,338 | 
| `convert length accumulator into single expression` | 6,329,138 | 1,051,338 |

raquo:
Cool, thanks for your work!

Issue #336, opened 2021-03-01
Swap out JSON parser for a faster implementation


lihaoyi:
uPickle's current JSON parser is forked from https://github.com/typelevel/jawn. While it's reasonably fast, parsing remains a significant portion of uPickle's execution time. 

The state of the art has since moved on, and faster JSON parsers have appeared:

- https://github.com/plokhotnyuk/jsoniter-scala
- https://github.com/sirthias/borer

We should fork one of the more modern JSON parsers so we can benefit from their JSON parsing performance. They should be able to plug directly into uPickle's Visitor-based architecture without any code changes on the uPickle side

uPickle's JSON renderer could probably be swapped out for a faster implementation as well

lihaoyi:
Some WIP in https://github.com/com-lihaoyi/upickle/tree/borer, which has the Borer JSON parser compiling, but more work is necessary to get it hooked up to uPickle's `Visitor`s

PR #337, opened 2021-03-02
Swap out forked Jawn parser for Borer parser

lihaoyi:
Attempt at https://github.com/com-lihaoyi/upickle/issues/336

More or less vendors https://github.com/sirthias/borer's `core/` folder verbatim, but eliding a bunch of things that we do not need e.g. `Encoder`/`Decoder`s and their surrounding infrastructure. Minor code changes were necessary to make things compile, and I changed the package names, but otherwise it was mostly a lift-and-shift.

A few tests needed to be tweaked because the error-reporting strategy for badly-tagged case class objects changed, from reporting the ending index of the object to reporting the start index. A handful of tests (3?) don't pass yet due to other error reporting changes, but most error reporting logic has been shimmed and works identically on both parsers

Most of the logic integrating Borer and uPickle lives in `ujson/src/ujson/BorerInputParser.scala`

Not really seeing much improvement in `./mill bench.jvm`:

Before:

```
RUN JVM: 500

benchmarkSampleJson 535001 bytes
benchmarkSampleMsgPack 412003 bytes
upickleDefaultCached Read 29
upickleDefaultCached Write 44
upickleLegacyCached Read 45
upickleLegacyCached Write 56

RUN JVM: 5000

upickleDefaultCached Read 608
upickleDefaultCached Write 747
upickleLegacyCached Read 729
upickleLegacyCached Write 727

RUN JVM: 50000

upickleDefaultCached Read 7144
upickleDefaultCached Write 8432
upickleLegacyCached Read 8220
upickleLegacyCached Write 8113
```

After

```
RUN JVM: 500

benchmarkSampleJson 535001 bytes
benchmarkSampleMsgPack 412003 bytes
upickleDefaultCached Read 9
upickleDefaultCached Write 26
upickleLegacyCached Read 15
upickleLegacyCached Write 37

RUN JVM: 5000

upickleDefaultCached Read 611
upickleDefaultCached Write 667
upickleLegacyCached Read 738
upickleLegacyCached Write 714

RUN JVM: 50000

upickleDefaultCached Read 7783
upickleDefaultCached Write 7967
upickleLegacyCached Read 9271
upickleLegacyCached Write 9384
```


lihaoyi:
Also seems to be a significant slowdown on `bench.js`:

Before:

```
RUN JS: 500

benchmarkSampleJson 535001 bytes
benchmarkSampleMsgPack 412003 bytes
upickleDefaultCached Read 25
upickleDefaultCached Write 7
upickleLegacyCached Read 20
upickleLegacyCached Write 8

RUN JS: 5000

upickleDefaultCached Read 277
upickleDefaultCached Write 78
upickleLegacyCached Read 252
upickleLegacyCached Write 80

RUN JS: 50000

upickleDefaultCached Read 2356
upickleDefaultCached Write 665
upickleLegacyCached Read 2317
upickleLegacyCached Write 762
```

After:

```
RUN JS: 500

benchmarkSampleJson 535001 bytes
benchmarkSampleMsgPack 412003 bytes
upickleDefaultCached Read 7
upickleDefaultCached Write 6
upickleLegacyCached Read 8
upickleLegacyCached Write 7

RUN JS: 5000

upickleDefaultCached Read 86
upickleDefaultCached Write 69
upickleLegacyCached Read 89
upickleLegacyCached Write 76

RUN JS: 50000

upickleDefaultCached Read 883
upickleDefaultCached Write 672
upickleLegacyCached Read 793
upickleLegacyCached Write 650
```

sirthias:
Unfortunately, I don't have time right now to look at the details of your port here, but there are two things that the borer parser needs in order to realize its full speed potential:

- read the UTF8-encoded bytes directly from the input byte arrays with `UNSAFE.getChar`, `UNSAFE.getInt` and `UNSAFE.getLong`
- specialize the parser for reading from bytes arrays via `Unsafe` with https://github.com/sirthias/borer/blob/master/project/src/main/scala/SpecializeJsonParserPlugin.scala

When these two things are done the parser really flies.
When they aren't, either because one is reading from some input other than byte arrays or `Unsafe` is not available (e.g. under scala-js), then parsing perf is unknown as I have never benchmarked these cases.
So, depending on your focus uses cases the borer parser might not actually be the best choice.

Concerning benchmarks:
I've found that artificial JSON benchmarks are all to often a trap. So I like to test perf against a range of larger real-world example files. This can mean a bit more work as one might have to connect the benchmarkees to several case class hierarchies but I find the results much more meaningful.

Here are the test files that I've used for [borer's benchmarks](https://github.com/sirthias/borer/tree/master/benchmarks/src/main/scala/io/bullet/borer/benchmarks) so far:
https://github.com/sirthias/borer/tree/master/benchmarks/src/main/resources

PR #340, opened 2021-03-07
Add Github Actions and publishVersion from git tag

lolgab:
- Remove Travis config file
- Update Mill to 0.9.5-52-ef6d5d


build.sc
```diff
@@ -1,35 +1,32 @@
...
+} yield (scalaV, scalaJSV)
 
-val scalaNativeVersions = Seq(
-  (scala212, scalaNative),
-  (scala213, scalaNative)
-)
+val scalaNativeVersions = for {
+  scalaV <- scala2Versions
+  scalaNativeV <- Seq("0.4.0")
+} yield (scalaV, scalaNativeV)
```
jodersky, line 29: Why change these? IMO the explicit version is easier to understand and follow

lolgab: Did it for consistency with the other repos. Reverted since this PR has another purpose either way.

PR #357, opened 2021-07-13
Fix: "match may not be exhaustive" scala 2.13 -Xlint warnings

### LINKED ISSUE #345: "match may not be exhaustive" in macroRW for case class
Hello, I was hoping to use [sbt-tpolecat](https://github.com/DavidGregory084/sbt-tpolecat) in a project of mine that depends on upickle, but it seems that macroRW is causing this compiler warning, so I can't make warnings fatal.

Minimized example: https://scastie.scala-lang.org/Pk7pFpgzT16Nccuyb7JKCA

```scala
import upickle.default._
case class Foo(num: Int)
val rw: ReadWriter[Foo] = macroRW
```

build.sbt
```
scalaVersion := "2.13.5"
scalacOptions ++= Seq("-Xlint",  "-Xfatal-warnings")
libraryDependencies += "com.lihaoyi" %% "upickle" % "1.3.8"
```

Result:

> match may not be exhaustive.
> It would fail on the following input: (x: Int forSome x not in 0)

htmldoug:
Fixes #345. Low-effort port of https://github.com/rallyhealth/weePickle/pull/82.

### DIFF
--- a/implicits/src-2/upickle/implicits/internal/Macros.scala
+++ b/implicits/src-2/upickle/implicits/internal/Macros.scala
@@ -259,6 +259,7 @@ object Macros {
                 for (i <- rawArgs.indices)
                 yield cq"$i => ${aggregates(i)} = v.asInstanceOf[${argTypes(i)}]"
               }
+              case _ => throw new java.lang.IndexOutOfBoundsException(currentIndex.toString)
             }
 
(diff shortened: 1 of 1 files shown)


lihaoyi:
@htmldoug lools good to me

PR #358, opened 2021-07-14
Implement `visitFloat32`, `visitFloat64String` for Byte/Short/Int/Long/Float/Double/Char

### LINKED ISSUE #354: MessagePack Float32 reading error
Reading `Float` values from *MessagePack* `Float32` values should be supported according to the [documentation](http://com-lihaoyi.github.io/upickle/#SupportedTypes) and it should be consistent with the writing behavior.

Currently the writing `Float` values creates a *MessagePack* `Float32` value but an equivalent reading operation causes the `FloatReader` to report an unexpected data type error.

### Environment
- uPickle: 1.4.0
- Scala: 3.0.0
- JRE: 17

### Reproduce
```scala
import upickle.default._

object Main extends App:

  println(readBinary[Float](writeMsg(1.0f)))
```

### Result
```scala
upickle.core.Abort: expected number got float32
	at upickle.core.SimpleVisitor.visitFloat32(SimpleVisitor.scala:29)
	at upickle.core.SimpleVisitor.visitFloat32$(SimpleVisitor.scala:7)
	at upickle.implicits.Readers$$anon$5.visitFloat32(Readers.scala:59)
	at upack.Msg$.transform(Msg.scala:182)
	at upack.Msg.transform(Msg.scala:23)
	at upack.Msg.transform$(Msg.scala:21)
	at upack.Float32.transform(Msg.scala:107)
	at upickle.Api.readBinary$$anonfun$1(Api.scala:30)
	at upickle.core.TraceVisitor$.withTrace(TraceVisitor.scala:18)
	at upickle.Api.readBinary(Api.scala:30)
	at upickle.Api.readBinary$(Api.scala:17)
	at upickle.default$.readBinary(Api.scala:133)
	at Main$.<clinit>(main.scala:6)
```

### Expected
```scala
1.0
```

### Fix
Add the following method override to the [FloatReader](https://github.com/com-lihaoyi/upickle/blob/master/implicits/src/upickle/implicits/Readers.scala#L66):
```scala
override def visitFloat32(d: Float, index: Int) = d
```

### Workaround
Add the following implicit reader override to your [custom configuration](https://com-lihaoyi.github.io/upickle/#CustomConfiguration):

```scala
  implicit override val FloatReader: Reader[Float] =
    new SimpleReader[Float] {

      override def expectedMsg = "expected number"
      override def visitString(s: CharSequence, index: Int) = s.toString.toFloat
      override def visitInt32(d: Int, index: Int) = d.toFloat
      override def visitInt64(d: Long, index: Int) = d.toFloat
      override def visitUInt64(d: Long, index: Int) = d.toFloat
      override def visitFloat32(d: Float, index: Int) = d
      override def visitFloat64(d: Double, index: Int) = d.toFloat
      override def visitFloat64StringParts(s: CharSequence, decIndex: Int, expIndex: Int, index: Int) =
        s.toString.toFloat
    }
```

htmldoug:
Fixes #354.

### DIFF
--- a/implicits/src/upickle/implicits/Readers.scala
+++ b/implicits/src/upickle/implicits/Readers.scala
@@ -42,7 +42,17 @@ trait Readers extends upickle.core.Types
     override def visitUInt64(d: Long, index: Int) = d
     override def visitFloat32(d: Float, index: Int) = d
     override def visitFloat64(d: Double, index: Int) = d
-
+    override def visitFloat64String(s: String, index: Int) = {
+      visitFloat64StringParts(
+        s,
+        s.indexOf('.'),
+        s.indexOf('E') match{
+          case -1 => s.indexOf('e')
+          case n => n
+        },
+        -1
+      )
+    }
     override def visitFloat64StringParts(s: CharSequence, decIndex: Int, expIndex: Int, index: Int) = {
       s.toString.toDouble
     }
@@ -54,6 +64,17 @@ trait Readers extends upickle.core.Types
     override def visitUInt64(d: Long, index: Int) = d.toInt
     override def visitFloat32(d: Float, index: Int) = d.toInt
     override def visitFloat64(d: Double, index: Int) = d.toInt
+    override def visitFloat64String(s: String, index: Int) = {
+      visitFloat64StringParts(
+        s,
+        s.indexOf('.'),
+        s.indexOf('E') match{
+          case -1 => s.indexOf('e')
+          case n => n
+        },
+        -1
+      )
+    }
     override def visitFloat64StringParts(s: CharSequence, decIndex: Int, expIndex: Int, index: Int) = {
       Util.parseIntegralNum(s, decIndex, expIndex, index).toInt
     }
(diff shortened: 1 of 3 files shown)


lihaoyi:
@htmldoug looks good, feel free to merge

lihaoyi:
@htmldoug could we add a unit test too? just to make sure this is fixing what we it should be fixing

htmldoug:
Sure. Tests found `visitFloat64String` missing, too. `JsVisitor` isn't on the classpath here, so I just copy-pasted. If that's okay with you, it's merge-ready.

lihaoyi:
looks good to me

lihaoyi:
Not sure if the copypaste code can easily be moved into some private helper function, but if it can that would help DRY things up a bit

htmldoug:
Refactored out a `NumericReader`. `protected` is the right visibility, I guess?

lihaoyi:
Seems good enough. This repo isn't particularly strict about visibility anyway. Feel free to merge

lihaoyi:
TBH I'd have put the logic into a static helper function somewhere that the various readers call, rather than a trait that the various readers extend. That seems like it would be a lot simpler, reduce the depth of the inheritance hierarchies, and make the logic easier to reuse if someone else wants to call it in odd places.

We do a lot of fancy trait cake stuff in the repo but I like static functions where we can get away with then

htmldoug:
I agree in principle, although it's a little tricky since they need to delegate to other instance methods.

Would either of these approaches be better?
```scala
def helper(s: String): (String, Int, Int) = ???

// each Reader:
def visitFloat64String(s: String, index: Int): T = {
  val (s, decIndex, expIndex) = helper(s)
  visitFloat64StringParts(s, decIndex, expIndex, index)
}
```

```scala
def helper[T](s: String, visitFloat64StringParts: (String, Int, Int) => T): T = ???

// each Reader:
def visitFloat64String(s: String, index: Int) = helper(s, visitFloat64StringParts(_, _, _, index))
```

lihaoyi:
Oh if a static helper is tricky lets just go with the trait then

PR #359, opened 2021-07-15
Run tests in CI again

lolgab:


.github/workflows/actions.yml
```diff
@@ -2,20 +2,54 @@ name: ci
...
+      - master
+    tags:
+      - '*'
   pull_request:
     branches:
       - master
 
 jobs:
-  test:
+  test-js:
```
lolgab, line 14: Here I split the jobs into js, native and jvm because the CI was taking something like 30 minutes to run!

.github/workflows/actions.yml
```diff
@@ -2,20 +2,54 @@ name: ci
...
+      JAVA_OPTS: "-Xss10M"
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-java@v1
+        with:
+          java-version: 8
+      - name: Run JS tests
+        run: |
+          ./mill -i __.resolvedIvyDeps
+          ./mill -i -j $(nproc) all __.js.__.compile __.js.__.publishArtifacts __.js.__.test
```
lolgab, line 26: Here I want to run in parallel because the CI is slow.
To avoid the Coursier problems when running in parallel I first run `__.resolvedIvyDeps` without `-j $(nproc)` so after that all the dependencies are already downloaded.
I tried to use `__.prepareOffline` instead of `__.resolvedIvyDeps` but it fails on `resolvedAmmoniteIvyDeps` for Scala js `0.6` and Scala `2.11`. Probably Ammonite dropped these versions long time ago.

build.sc
```diff
@@ -38,7 +39,11 @@ val scalaNativeVersions = Seq(
 )
 
 trait CommonModule extends ScalaModule {
-  def scalacOptions = T{ if (scalaVersion() == scala212) Seq("-opt:l:method") else Nil }
+  def scalacOptions = T{
+    super.scalacOptions() ++ {
```
lolgab, line 43: To avoid loosing the `-scalajs` option, I added  `super.scalacOptions() ++` everywhere.

build.sc
```diff
@@ -49,7 +54,14 @@ trait CommonModule extends ScalaModule {
...
       Seq(PathRef(millSourcePath / "src-2.12+"))
-    } else Seq())
+    } else Seq()) ++
+    (if (scalaVersion() == scala212 || scalaVersion() == scala213) {
+      Seq(PathRef(millSourcePath / "src-2.12-2.13"))
+    } else Seq()) ++
+    (platformSegment match {
+      case "jvm" | "native" => Seq(PathRef(millSourcePath / "src-jvm-native"))
+      case _ => Seq()
+    })
```
lolgab, line 64: Moved some tests in some specific places

build.sc
```diff
@@ -82,32 +94,32 @@ trait CommonPublishModule extends CommonModule with PublishModule with CrossScal
...
       mill.modules.Jvm.createJar(Agg(javadocDir))(outDir)
-    } else T {
-      super.docJar()
+    } else {
+      super.docJar
     }
   }
 }
 
+trait CommonTestModule extends CommonModule with TestModule{
```
lolgab, line 109: Moved all the traits to the root module, thinking this was causing problems.

build.sc
```diff
@@ -119,6 +131,7 @@ trait CommonJsModule extends CommonPublishModule with ScalaJSModule{
   def platformSegment = "js"
   def crossScalaJSVersion: String
   def scalaJSVersion = crossScalaJSVersion
+  def scalacOptions = super.scalacOptions()
```
lolgab, line 134: I don't know why, but if I remove this line some modules loose the `-scalajs` option.
Couldn't find where this happens since there is no redefinition of `scalacOptions` without `super.scalacOptions() ++`

PR #361, opened 2021-08-11
BugFix: Util.parseLong

htmldoug:
Currently:
```scala
upickle.default.read[Long](""""a"""") ==> 49L
```

Port of https://github.com/rallyhealth/weePickle/pull/88.

### DIFF
--- a/core/src/upickle/core/Util.scala
+++ b/core/src/upickle/core/Util.scala
@@ -63,7 +63,7 @@ object Util {
 
     intPortion + decPortion
   }
-  def parseLong(cs: CharSequence, start: Int, len: Int): Long = {
+  def parseLong(cs: CharSequence, start: Int, end: Int): Long = {
 
     // we store the inverse of the positive sum, to ensure we don't
     // incorrectly overflow on Long.MinValue. for positive numbers
@@ -71,9 +71,8 @@ object Util {
     var inverseSum: Long = 0L
     var inverseSign: Long = -1L
     var i: Int = start
-    val end = start + len
 
-    if ((start | len | (cs.length - end)) < 0) throw new IndexOutOfBoundsException
+    if ((start | end | (cs.length - end)) < 0) throw new IndexOutOfBoundsException
 
     if (cs.charAt(start) == '-') {
       inverseSign = 1L
--- a/core/src/upickle/core/Util.scala
+++ b/core/src/upickle/core/Util.scala
@@ -72,7 +72,7 @@ object Util {
     var inverseSign: Long = -1L
     var i: Int = start
 
-    if ((start | end | (cs.length - end)) < 0) throw new IndexOutOfBoundsException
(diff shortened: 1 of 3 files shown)


htmldoug:
That first commit should have failed in CI. Seems like tests aren't being run. Maybe related to #359.

lihaoyi (approved):

lolgab:
Rebased this PR after merging https://github.com/com-lihaoyi/upickle/pull/359 which fixes the CI.

lolgab:
@htmldoug Can you check the failing test?

htmldoug:
Thanks so much for fixing up the build and tests. Good thing too, since my assumptions about this method were wrong.

Should be good now, although I've jumped around with approaches and pickle flavors so much that I'd appreciate another set of eyes on it before merging.

core/src/upickle/core/Util.scala
```diff
@@ -72,25 +72,26 @@ object Util {
     var inverseSign: Long = -1L
     var i: Int = start
 
+    if ((start | end | end - start | (cs.length - end)) < 0) throw new IndexOutOfBoundsException
```
lolgab, line 75: Sorry, I can't understand this check here.
You make the or of the `start`, `end` etc. and check if the result is `< 0`. Why?
Also are you sure there isn't redundancy between this check and  the one in line 83?

lolgab: I think I got the meaning. You do the OR of all these numbers since none of them should have the sign bit at `1`. So this is equivalent to:
```scala
if (start < 0 || end < 0  || (end - start) < 0 || (cs.length - end) < 0) throw new IndexOutOfBoundsException
```
About the redundancy of the checks, the one in line 83 (`size <= 0`) checks if you're trying to parse just a `"-"` which should fail. Right?

lolgab:
I'm approving and merging the PR as it is.
Any improvements can be done later as a follow-up PR.

core/src/upickle/core/Util.scala
htmldoug: Right. The `size <= 0` check should result in an exception thrown for `"-"`.

Credit to [java.io.FilterOutputStream](https://github.com/openjdk/jdk/blob/cec6c068b03d890312e50b448fbc26102c635249/src/java.base/share/classes/java/io/FilterOutputStream.java#L133) for the arcane bitwise bounds check. I assume it was written that way for performance, but I haven't measured if the difference is actually noticeable over your equivalent and clearer version. That class says `@since JDK1.0` which was 1996. CPU branch prediction and the JVM implementation have improved since then.

The bounds check isn't really related to the fix, but since I found one bug and didn't have a good understanding of how this is used, I wanted to nail down the assumptions around inputs. I think `parseLong` is used deep in the parsing code (not public API) so the check could probably be removed if needed, although I didn't expect the cost of keeping it to be significant.

htmldoug:
Thanks for the review and merge, @lolgab!

PR #364, opened 2021-09-05
Check backwards binary compatibility with Mima

lolgab:


julienrf:
Thank you @lolgab, this is a great step forward!

Do you also plan to follow the [recommended versioning scheme](https://docs.scala-lang.org/overviews/core/binary-compatibility-for-library-authors.html#recommended-versioning-scheme)? If so, do you mind declaring the project versioning scheme as `properties` metadata in the pom.xml?

~~~ xml
<info.versionScheme>early-semver</info.versionScheme>
~~~

lolgab:
Hi @julienrf,
Thank you! I really believe that a better versioning scheme is what this ecosystem needs most to thrive!
However, we didn't formalize yet what versioning scheme we're going to follow. Once we do and we make sure that libraries actually follow it, we can add the versionScheme field in the pom.
Actually, I think this first needs to be supported in `mill` to be able to add the field in first place :D ( I'll open a PR for that! )

julienrf:
I agree it would be great if mill offered first-class support for declaring the versioning scheme. Let me know if I can help!

Issue #360, opened 2021-07-27
Default values not working with @key annotation in Scala 3


pharod:
```scala 
case class Person(
                     @key("first_name") firstName: String = "N/A",
                     @key("last_name") lastName: String)
object Person {
    implicit val rw: RW[Person] = macroRW
}

val json = """{"last_name": "Snow"}"""
val p = read[Person](json)
```
This code fragment executes with exception:

```java
Exception in thread "main" java.lang.ExceptionInInitializerError
	at com.nitka.rb.RBSyncApp.main(RBSyncApp.scala)
Caused by: upickle.core.AbortException: missing keys in dictionary: first_name at index 20
	at ujson.CharParser$$anon$1.applyOrElse(CharParser.scala:343)
	at ujson.CharParser$$anon$1.applyOrElse(CharParser.scala:341)
	at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:35)
	at ujson.CharParser.liftedTree1$1(CharParser.scala:496)
	at ujson.CharParser.tryCloseCollection(CharParser.scala:496)
	at ujson.CharParser.parseNested(CharParser.scala:462)
	at ujson.CharParser.parseTopLevel0(CharParser.scala:323)
	at ujson.CharParser.parseTopLevel(CharParser.scala:307)
	at ujson.CharParser.parse(CharParser.scala:59)
	at ujson.StringParser$.transform(StringParser.scala:28)
	at ujson.StringParser$.transform(StringParser.scala:28)
	at ujson.Readable$fromTransformer.transform(Readable.scala:13)
	at upickle.Api.read$$anonfun$1(Api.scala:37)
	at upickle.core.TraceVisitor$.withTrace(TraceVisitor.scala:18)
	at upickle.Api.read(Api.scala:37)
	at upickle.Api.read$(Api.scala:17)
	at upickle.default$.read(Api.scala:133)
	at com.nitka.rb.RBSyncApp$.<clinit>(RBSyncApp.scala:76)
	... 1 more
Caused by: upickle.core.Abort: missing keys in dictionary: first_name
	at com.nitka.rb.RBSyncApp$$anon$1.make(RBSyncApp.scala:72)
	at com.nitka.rb.RBSyncApp$$anon$1.make(RBSyncApp.scala:72)
	at upickle.implicits.CaseClassReaderPiece$$anon$1.visitEnd(CaseClassReader.scala:30)
	... 16 more
``` 
[...]

PR #365, opened 2021-09-30
Fix bugs in scala3 code #362 #360

### LINKED ISSUE #362: Using Scala 3 seriously breaks default value handing
Minimal example:

```scala
case class A(a: Int = 1)

import upickle.default._
implicit val rw: ReadWriter[A] = macroRW[A]

println(read[A]("{\"a\":2}"))
println(read[A]("{}"))
```

Expected (upickle 1.4.0, Scala 2.13.6):
```
A(2)
A(1)
```

Got (upickle 1.4.0, Scala 3.0.1):
```
A(2)
A(2)
```

God knows how many projects are silently corrupting their databases. Or leaking data from another user in their API calls. Since the parser is keeping and leaking state, everything is possible.

lihaoyi:
Fixes #362 #360

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -75,7 +75,7 @@ trait CommonPublishModule extends CommonModule with PublishModule with Mima with
       .lastTag
       .getOrElse(throw new Exception("Missing last tag"))
 
-    Seq("1.4.0", lastTag)
+    if (isDotty) Nil else Seq("1.4.0", lastTag)
   }
   def isDotty = crossScalaVersion.startsWith("0") || crossScalaVersion.startsWith("3")
   def pomSettings = PomSettings(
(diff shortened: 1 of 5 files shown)


lihaoyi:
@jodersky could I get a review

lihaoyi:
Going to merge this; not sure what the binary compatibility complaint is, since this only touches scala 3 stuff

lihaoyi:
Seems like the bincompat complaint is MIMA not liking the scala3 version bump. Just disabling MIMA for Scala3 for now

lolgab:
Great @lihaoyi!
I'll check what's going on with MiMa when I have some time :)

jodersky:
looks good, thanks for the fix!

PR #366, opened 2021-09-30
Use properly decoded field names when handling case classes

lihaoyi:


lihaoyi:
@htmldoug could I ask you for a review on this? :)

lihaoyi:
Gonna merge this; the existing tests should make sure nothing broke, and the new test should ensure the issue is fixed

PR #370, opened 2021-11-07
MsgPackReader: properly increment index in `ext`

### LINKED ISSUE #369: upack doesn't roundtrip `Ext` nested inside another `Msg`
Here's a minimal roundtripping failure:

```scala
val msg = upack.Arr(upack.Ext(33, new Array[Byte](4)), upack.False)
val bytes = upack.writeToByteArray(msg)
val msg2 = upack.read(bytes)
```

This decodes as `upack.Arr(upack.Ext(33, new Array[Byte](4)), upack.Int32(33))` (the last bit is the problem: `upack.Int32(33)` instead of `upack.False`).

harpocrates:
The `parseExt` helper in `MsgPackReader` was not properly updating the
position in the data source.

Added two regression tests. These compare serialized and re-serialized
values (instead of just initial and serialized) since structurally identical `Ext`
messages won't compare identical (due to the `Array[Byte]` field whose
equality is determined by identity).

Fixes #369

### DIFF
--- a/upack/src/upack/Msg.scala
+++ b/upack/src/upack/Msg.scala
@@ -128,7 +128,13 @@ object Obj{
 
   def apply(): Obj = Obj(new mutable.LinkedHashMap[Msg, Msg]())
 }
-case class Ext(tag: Byte, data: Array[Byte]) extends Msg
+case class Ext(tag: Byte, data: Array[Byte]) extends Msg {
+  override def equals(other: Any): Boolean = other match {
+    case Ext(tagOther, dataOther) =>
+      tag == tagOther && java.util.Arrays.equals(data, dataOther)
+    case _ => false
+  }
+}
 
 sealed abstract class Bool extends Msg{
   def value: Boolean
--- a/upack/src/upack/Msg.scala
+++ b/upack/src/upack/Msg.scala
@@ -128,16 +128,7 @@ object Obj{
 
   def apply(): Obj = Obj(new mutable.LinkedHashMap[Msg, Msg]())
 }
-case class Ext(tag: Byte, data: Array[Byte]) extends Msg {
-  override def equals(other: Any): Boolean = other match {
-    case Ext(tagOther, dataOther) =>
-      tag == tagOther && java.util.Arrays.equals(data, dataOther)
-    case _ => false
-  }
-
-  override def hashCode: Int =
-    MurmurHash3.bytesHash(b, "Ext".hashCode)
-}
+case class Ext(tag: Byte, data: Array[Byte]) extends Msg
 
 sealed abstract class Bool extends Msg{
   def value: Boolean
--- a/upack/src/upack/Msg.scala
+++ b/upack/src/upack/Msg.scala
@@ -134,6 +134,9 @@ case class Ext(tag: Byte, data: Array[Byte]) extends Msg {
       tag == tagOther && java.util.Arrays.equals(data, dataOther)
     case _ => false
   }
+
+  override def hashCode: Int =
+    MurmurHash3.bytesHash(b, "Ext".hashCode)
 }
 
 sealed abstract class Bool extends Msg{
--- a/upack/src/upack/MsgPackReader.scala
+++ b/upack/src/upack/MsgPackReader.scala
@@ -98,7 +98,9 @@ abstract class BaseMsgPackReader extends upickle.core.BufferingByteParser{
   }
   def parseExt[T](n: Int, visitor: Visitor[_, T]) = {
     val (arr, i, j) = sliceArr(index + 1, n)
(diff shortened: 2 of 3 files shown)


upack/src/upack/Msg.scala
```diff
@@ -128,7 +128,13 @@ object Obj{
 
   def apply(): Obj = Obj(new mutable.LinkedHashMap[Msg, Msg]())
 }
-case class Ext(tag: Byte, data: Array[Byte]) extends Msg
+case class Ext(tag: Byte, data: Array[Byte]) extends Msg {
+  override def equals(other: Any): Boolean = other match {
```
htmldoug, line 132: Custom `hashCode()` would be a good idea to cover the rare but possible case that someone uses `Ext` as a Map key.

https://stackoverflow.com/a/2265637

harpocrates: Oh bother, of course.

harpocrates: Done now.

upack/src/upack/Msg.scala
```diff
@@ -128,7 +128,16 @@ object Obj{
...
+    case Ext(tagOther, dataOther) =>
+      tag == tagOther && java.util.Arrays.equals(data, dataOther)
+    case _ => false
+  }
+
+  override def hashCode: Int =
+    MurmurHash3.bytesHash(b, "Ext".hashCode)
```
htmldoug, line 139: `tag` should be represented in the hash somehow. We could probably just use it as the seed.

upack/src/upack/MsgPackReader.scala
```diff
@@ -98,7 +98,9 @@ abstract class BaseMsgPackReader extends upickle.core.BufferingByteParser{
   }
   def parseExt[T](n: Int, visitor: Visitor[_, T]) = {
     val (arr, i, j) = sliceArr(index + 1, n)
-    visitor.visitExt(getByteSafe(index), arr, i, j, index)
+    val res = visitor.visitExt(getByteSafe(index), arr, i, j, index)
+    index += n + 1
```
htmldoug, line 102: This looks important!

If I'm reading correctly, looks like `n` is the array length, and the `+ 1` is the tag/ext-type byte. Seems like this fix should cover the fixext types as well. 👍

htmldoug (approved):
lgtm. thanks!

upack/src/upack/Msg.scala
harpocrates: The more I think about this, the less comfortable I am about updating `equals` and `hashCode` here. That could lead to some end-user changes and it isn't consistent with some of the other types (eg. `Bytes`). Maybe this is still desirable change, but probably not in a PR about fixing ext parsing.

I'm going to switch the round-tripping test I added to assert equivalence of the serialized and serialized-deserialized-reserialized payloads (that way we don't need to call `equals` on a `Msg`)

htmldoug: 👍 for smaller, well-scoped PRs.

Re: `equals`/`hashCode`, it's hard to imagine that any users are depending on the behavior being broken other than to work around it.

Good point about `case class Binary(value: Array[Byte]) extends Msg` being affected, too. A fix for one should probably fix both.

htmldoug: The rabbit hole goes deeper. Given that `0.0d == 0L`, should `Float64(0.0) == Int64(0L)`? Probably so! This is especially relevant since round-tripping `Int64(0L)` through messagepack will come back as `Int32(0)`. I think you're wise to push `equals`/`hashCode` to a separate PR. Making assertions against the `Array[Byte]` sounds good for now.

harpocrates:
I've switched back to `Array[Byte]` for the tests.

If you are planning on squashing and merging, maybe use that as the commit message instead of the default concatenation of intermediate messages?

upack/src/upack/Msg.scala
harpocrates: > Given that `0.0d == 0L`, should `Float64(0.0) == Int64(0L)`? 

Personally, I don't think this should be the case - the "int format family" and "float format family" are separate. I further wish that `0.0d == 0L` would fail to typecheck instead of returning true (that's another rabbit whole though, and Scala is unlikely yo ever change that).

> This is especially relevant since round-tripping `Int64(0L)` through messagepack will come back as `Int32(0)`.

I agree `Int64(0L)` should roundtrip, but on the other hand: is there really a need for the distinction? Why not just a single `case class Int(value: Long) extends Msg` (and `UInt` for unsigned)? We don't distinguish `Int16` or the fixint cases, so why distinguish the 32 and 64 bit cases?

htmldoug (approved):

Issue #356, opened 2021-06-23
Writing Scala 3 enums is broken


lolgab:
Trying to write Scala 3 `enum` values actually writes always the first one.

### Test reproducing the problem

```scala
package upickle

import scala.language.implicitConversions
import utest._

import upickle.default._

object EnumTests extends TestSuite {

  enum SimpleEnum {
    case A, B
  }

  val tests = Tests {
    test("simple") {
      given ReadWriter[SimpleEnum] = macroRW[SimpleEnum]
      test("enum write") {
        val parsed = write(SimpleEnum.B)
        val expected = """{"$type":"upickle.EnumTests.SimpleEnum.B"}"""
        assert(parsed == expected)
      }
    }
  }
}

```

Fails with:
```
X upickle.EnumTests.simple.enum write 71ms 
  utest.AssertionError: parsed == expected
  parsed: String = {"$type":"upickle.EnumTests.SimpleEnum.A"}
  expected: String = {"$type":"upickle.EnumTests.SimpleEnum.B"}
```

htmldoug:
There might be something useful that can be ported over from https://github.com/rallyhealth/weePickle/pull/94.

Lasering:
```scala
enum Color derives Writer:
  case Red, Green, Blue
println(write(Color.Blue))
    println(write(Color.Red))
    println(write(Color.Green))
```

Prints `{"$type":"work.martins.simon.simulaev.Color.Red"}` three times.

Lasering:
The problem is most likely in the [macros.summonList](https://github.com/com-lihaoyi/upickle/blob/c6eacaf4cf296f7638f17b89541938101a71f52d/implicits/src-3/upickle/implicits/macros.scala#L34). It should be something like:

```
inline def summonList[T <: Tuple]: List[_] =
  inline erasedValue[T] match
    case _: EmptyTuple => Nil
    case _: (t *: ts) => summonInline[t] :: summonList[ts]
end summonList

inline final def summonWriters[T <: Tuple]: List[Writer[_]] =
  inline erasedValue[T] match
    case _: EmptyTuple => Nil
    case _: (t *: ts) => summonWriter[t] :: summonWriters[ts]

inline final def summonWriter[A]: Writer[A] =
  summonFrom {
    case w: Writer[A] => w
    case _: Mirror.Of[A] => Writer.derived[A]
  }
```

BoopBoopBeepBoop:
[...]

PR #374, opened 2021-12-19
Bugfix: Scala 3 enums (#356)

BoopBoopBeepBoop:
Adds support for scala 3 enumerations which reads/writes them as their canonical string representations (used by the builtin "valueOf" method).

### DIFF
--- a/implicits/src-3/upickle/implicits/CaseClassReader.scala
+++ b/implicits/src-3/upickle/implicits/CaseClassReader.scala
@@ -1,8 +1,12 @@
 package upickle.implicits
 
-import compiletime.{summonInline}
+import compiletime.summonInline
 import deriving.Mirror
-import upickle.core.{ Visitor, ObjVisitor, Annotator }
+import upickle.core.{Annotator, ObjVisitor, Visitor, Abort}
+import upickle.implicits.macros.EnumDescription
+
+import deriving.*
+import compiletime.*
 
 trait CaseClassReaderPiece extends MacrosCommon:
   this: upickle.core.Types with Readers with Annotator =>
@@ -29,6 +33,19 @@ trait CaseClassReaderPiece extends MacrosCommon:
         make(builder.toMap)
     }
   end CaseClassReader
+
+  class EnumReader[T](f: String => T, description: EnumDescription) extends SimpleReader[T]:
+    override def expectedMsg = "expected string enumeration"
+    override def visitString(s: CharSequence, index: Int) = {
+      val str = s.toString
+      try {
+        f(str)
+      } catch {
+        case _: IllegalArgumentException =>
+          throw new Abort(s"Value '$str' was not found in enumeration ${description.pretty}")
+      }
+    }
+  end EnumReader
 
   inline def macroR[T](using m: Mirror.Of[T]): Reader[T] = inline m match {
     case m: Mirror.ProductOf[T] =>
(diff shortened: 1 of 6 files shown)


lihaoyi:
@jodersky not sure if I could bother you to look at this one as well :)

implicits/src-3/upickle/implicits/CaseClassReader.scala
```diff
@@ -1,8 +1,12 @@
...
 
-import compiletime.{summonInline}
+import compiletime.summonInline
 import deriving.Mirror
-import upickle.core.{ Visitor, ObjVisitor, Annotator }
+import upickle.core.{Annotator, ObjVisitor, Visitor, Abort}
+import upickle.implicits.macros.EnumDescription
+
+import deriving.*
+import compiletime.*
```
jodersky, line 9: unless members of these are truly used everywhere in this file and it is clear from context, please avoid using wildcard imports

implicits/src-3/upickle/implicits/CaseClassReader.scala
```diff
@@ -82,11 +99,27 @@ trait CaseClassReaderPiece extends MacrosCommon:
...
 
+  inline def macroEnumR[T](using m: Mirror.Of[T]): Reader[T] = inline m match {
+    case m: Mirror.ProductOf[T] =>
+      throw new UnsupportedOperationException("Generated Enum Writer should never encounter Product")
+
+    case m: Mirror.SumOf[T] =>
+      val valueOf = macros.enumValueOf[T]
+      val description = macros.enumDescription[T]
+      new EnumReader[T](valueOf, description)
+  }
```
jodersky, line 110: For backwards compatibility, I think it would be useful to have this method folded into in the `macroR` method.This could be merged as a case in the general `macroR` method. E.g.

```scala
inline def macroR[T](using m: Mirror.Of[T]): Reader[T] = inline m match {
  case m: Mirror.ProductOf[T] => ...
  case m: SumOf[T] if erasedValue[T].isInstanceOf[scala.reflect.Enum] => <content of macroEnumR>
  case m: SumOf[T] => ...
}
```

The reason I say that this is for backwards compatibility, is that there is lot of scala 2 code out there defines implicit ReadWriters calling to the macroR method, and it would be nice if we would get new enum functionality in those cases too, even if the old syntax is still kept around.

implicits/src-3/upickle/implicits/CaseClassReader.scala
```diff
@@ -82,11 +99,27 @@ trait CaseClassReaderPiece extends MacrosCommon:
...
+
   inline given [T <: Singleton: Mirror.Of]: Reader[T] = macroR[T]
 
   // see comment in MacroImplicits as to why Dotty's extension methods aren't used here
   implicit class ReaderExtension(r: Reader.type):
-    inline def derived[T](using Mirror.Of[T]): Reader[T] = macroR[T]
+    inline def derived[T](using Mirror.Of[T]): Reader[T] = inline erasedValue[T] match {
+      case _: scala.reflect.Enum =>
```
jodersky, line 117: as mentioned in my comment above, for backwards-compatibility/upgradability, I think that the derivation logic should all be contained in the macroR method, and this should be a simple forwarder to it

implicits/src-3/upickle/implicits/CaseClassWriter.scala
```diff
@@ -4,6 +4,7 @@ import compiletime.{summonInline}
 import deriving.Mirror
 import scala.reflect.ClassTag
 import upickle.core.{ Visitor, ObjVisitor, Annotator }
+import deriving.*, compiletime.*
```
jodersky, line 7: please avoid wildcard imports here too

implicits/src-3/upickle/implicits/CaseClassWriter.scala
```diff
@@ -59,11 +64,23 @@ trait CaseClassWriterPiece extends MacrosCommon:
       Writer.merge[T](writers:_*)
   }
 
+  inline def macroEnumW[T](using m: Mirror.Of[T]): Writer[T] = inline m match {
```
jodersky, line 67: same comment about folding macroEnumW into macroW, as I mentioned about the reader case

jodersky (commented):
Thanks for this contribution. The core looks very good, however I have comments on how to restructure some of the methods, so that users can reap the benefits when working with (partial) scala 2 codebases

jodersky:
Moved to #378. @BoopBoopBeepBoop, feel free to open a new PR if you have more amendments to the updated PR.

Issue #353, opened 2021-06-09
Generic case class error in Scala 3


martin-ockajak:
Case classes with type parameters should be supported according to the [documentation](http://com-lihaoyi.github.io/upickle/#SupportedTypes).

Currently this does not seem to work and causes the case class reader object traversal to crash.

### Environment
- uPickle: 1.4.0
- Scala: 3.0.0
- JRE: 17

### Reproduce
```scala
import upickle.default._

object Main extends App:

  case class Container[T](
    value: T
  )

  given ReadWriter[Container[String]] = macroRW

  val container = Container("test")
  val json = """{ "value": "test" }"""

  println("\nWrite:")
  println(write(container))

  println("\nRead:")
  println(read[Container[String]](json))
```

### Result
```scala
Write:
{}

Read:
java.lang.ArrayIndexOutOfBoundsException: 0
	at Main$$anon$2.productElement(main.scala:9)
	at Main$Container$.fromProduct(main.scala:5)
	at Main$Container$.fromProduct(main.scala:5)
	at Main$$anon$1.make(main.scala:9)
	at Main$$anon$1.make(main.scala:9)
	at upickle.implicits.CaseClassReaderPiece$$anon$1.visitEnd(CaseClassReader.scala:30)
	at ujson.CharParser.liftedTree1$1(CharParser.scala:496)
	at ujson.CharParser.tryCloseCollection(CharParser.scala:496)
	at ujson.CharParser.parseNested(CharParser.scala:462)
	at ujson.CharParser.parseTopLevel0(CharParser.scala:323)
	at ujson.CharParser.parseTopLevel(CharParser.scala:307)
	at ujson.CharParser.parse(CharParser.scala:59)
	at ujson.StringParser$.transform(StringParser.scala:28)
	at ujson.StringParser$.transform(StringParser.scala:28)
	at ujson.Readable$fromTransformer.transform(Readable.scala:13)
	at upickle.Api.read$$anonfun$1(Api.scala:37)
	at upickle.core.TraceVisitor$.withTrace(TraceVisitor.scala:18)
	at upickle.Api.read(Api.scala:37)
	at upickle.Api.read$(Api.scala:17)
	at upickle.default$.read(Api.scala:133)
	at Main$.<clinit>(main.scala:18)
```

### Expected
```scala
Write:
{ "value" : "test" }

Read:
Container("test")
```

martin-ockajak:
[...]

PR #377, opened 2021-12-30
Bugfix: Scala 3 generics (#353)

mgyucht:
A small change to `fieldLabelsImpl0` seems to be enough to fix `macroRW` for classes with type parameters in Scala 3. I've added a small regression test as well. 

In the existing implementation, `TypeTree.of[T].symbol` evaluates to `noSymbol` when using a parameterized type. `TypeRepr.of[T].typeSymbol.primaryConstructor.paramSymss` seems to expose the full parameter list including type parameters, so we filter out any type parameters and are left with the constructor fields.

I'm not very confident with compiler internals, in particular not 100% sure I understand the ramifications of using TypeTree vs TypeRepr, so I would appreciate a careful review of this change.


lihaoyi:
@jodersky passing this along to you since you know the scala3 stuff :)

jodersky (approved):
Looks good, thanks

PR #378, opened 2022-01-03
Bugfix: Scala 3 enums (#356)

jodersky:
I ammended @BoopBoopBeepBoop's PR #374 with the proposed fixes.

@lihaoyi, the code itself LGTM, however I realized that we never discussed the functionality. This will change the serialization of Scala 3 enums from using tagged union serialization to simple strings. For instance, 

```
enum SimpleEnum derives ReadWriter:
  case A, B

write(SimpleEnum.A)
```

is currently serialized as:

```
{"$type":"upickle.SimpleEnum.A"}
```

and will now be serialized as:

```
"A"
```

is this something that is desirable? There are a couple of potential issues that could arise:
- More potential for ambiguity and/or name clashes in deserialization. Often we know what type is expected when deserializing something, but I could imagine that there are situations where that isn't the case.
- Enums can no longer be returned as top-level JSON structures without violating [the RFC](https://datatracker.ietf.org/doc/html/rfc4627), which allows only JSON objects and arrays as top-level structures. I don't know the severity of this issue in practice, but it technically breaks the spec.


lihaoyi:
TBH I'd rather we fix the bug but keep the existing `{"$type": "..."}`-based serialization format intact, and have a separate PR where we add a flag to let people enable the more concise serialization format. 

In the follow up, we should make the serialization format change opt-in to avoid breakage, and make sure the concise serialization format applies to `case object`s too for consistency.

BoopBoopBeepBoop:
Sorry that I was taking a while on revising - @jodersky it was precisely because I was attempting to address what you're asking about (the functionality change), and that took me down the rabbit hole of needing to alter the tagged writers specifically for scala 3, and then I believe I got blocked at the point where it seemed that with the removal of the `ClassTag` mechanism I would need to do something like "generate a match statement containing all of the types below this sum type", which stumped me. 

I was not able to easily find a way to do that via the compile time API's - if you have pointers on this I'd love to know how to achieve it. It appeared to be necessary because with the classtag removal, you need to have a way to choose a child writer in a hierarchy from an instance of that object (hence "generate the case statement"). It seemed like somehow erasedValue would allow you to do it, but couldn't work it out.

PR #380, opened 2022-01-25
Support Scala Native - Scala 3

lolgab:
Bump Geny and Utest

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Run JS tests
         run: |
           ./mill -i __.resolvedIvyDeps
-          ./mill -i -j $(nproc) all __.js.__.compile __.js.__.publishArtifacts __.js.__.test
+          ./mill -i -j 0 all __.js.__.compile __.js.__.publishArtifacts __.js.__.test
   test-native:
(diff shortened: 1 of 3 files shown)


lolgab:
Since the libraries were updated to a binary incompatible version, I would bump the minor version. What do you think?

jodersky:
Nice! I recently found out that Scala 3.1 has a new `-Yrelease` flag, which allows setting which minor version it will target. If it makes it easier, we could use Scala 3.1 (and newer versions in the future) everywhere in the build and set the flag to the lowest appropriate version.

jodersky:
oh wait, that flag is only available from Scala 3.1.2-RC1 onwards

jodersky (approved):

lolgab:
Yes, when `3.1.2` is out we can migrate the builds to do that :)

lolgab:
@jodersky About the minor tagging. Does it sound like a good middle ground?

jodersky:
I am fine with that

jodersky:
actually wait one sec. What libraries were updated? Acyclic is only part of the tests, no?

jodersky:
ah no, geny was updated too. Let's go for it then

PR #382, opened 2022-03-07
Serialize case objects as literal strings

lihaoyi:
Now, `case object Foo` is now serialized as `"pkg.path.Foo"`, rather than `{"$type": "pkg.path.Foo"}`. This simplifies the output, and makes serialized `sealed trait`/`case object` hierarchies look much more like simple enumerations when serialized. This harmonizes them with the simple serialization of Scala `enum`s in  https://github.com/com-lihaoyi/upickle/pull/378.

`case class`es, even with zero parameters, are still serialized using the `{"$type": "pkg.path.Foo"}` encoding.

This is a breaking change; uPickle with this change continues to be able to read the old format, but older versions of uPickle will not be able to read the new format. This will need to go into uPickle 2.0.0, with appropriate migration support to smooth the transition


lihaoyi:
@jodersky `upickle.MacroTests.commonCustomStructures.singleton` fails on this PR on Scala 3. It seems that the Scala 3 macro implementation does not generate `SingletonW`/`SingletonR`s like the Scala 2 implementation does. Any idea how I can make it do that?

PR #395, opened 2022-07-04
deps: update a bunch of things

ckipp01:
This updates the following:

  - make sure Mill is on the latest 0.10.5
  - drops support for scalajs 0.x
  - bumps to the latest scala patch versions
  - bumps to the latest scalaJS and scala native
  - bumps to 0.8.0 of utest
  - adds in dependabot
  - adds in cross testing jdk for 8 and 17
  - bumps version of checkout and java actions


ckipp01:
Alright so this repo is a bit more involved and has more deps in various modules so instead of bumping everything at once, I think this is a good one to just get this taken care of, add it to steward, and then let them come in one at a time. It'll be easier to catch regressions in a single dep that way.

build.sc
```diff
@@ -8,39 +8,26 @@ import mill.scalalib.api.Util.isScala3
...
-val scala30   = "3.0.2"
-val scala31   = "3.1.0"
-val scalaJS06 = "0.6.33"
-val scalaJS1  = "1.5.1"
-val scalaNative = "0.4.3"
+val scala212  = "2.12.16"
+val scala213  = "2.13.8"
+val scala3   = "3.1.3"
+val scalaNative = "0.4.5"
+val scalaJS1 = "1.10.1"
```
lolgab, line 19: Since there is only one ScalaJS now, it can be named without `1`

lolgab:
Thank you for this @ckipp01 ? :-)

Issue #433, opened 2023-01-20
Using writeJs with Map of anything but String produces error


mgilbertnz:
Testing on Version 2.0.0

`write` works correctly.
`writeJs` only works if `Map` is `Map[String, String]`

```
import upickle.default._
writeJs(Map("hello1" -> "hello2")) // Works
writeJs(Map("hello1" -> 1)) // Error
writeJs(Map(1 -> "hello2")) // Error
```

```
upickle.core.Abort: expected string got int32
	at upickle.core.SimpleVisitor.visitInt32(SimpleVisitor.scala:31)
	at upickle.core.SimpleVisitor.visitInt32$(SimpleVisitor.scala:31)
	at upickle.core.StringVisitor$.visitInt32(StringVisitor.scala:3)
	at upickle.implicits.Writers$$anon$4.write0(Writers.scala:30)
	at upickle.implicits.Writers$$anon$4.write0(Writers.scala:28)
	at upickle.core.Types$Writer.write(Types.scala:130)
	at upickle.core.Types$Writer.write$(Types.scala:128)
	at upickle.implicits.Writers$$anon$4.write(Writers.scala:28)
	at upickle.implicits.Writers$$anon$18.$anonfun$write0$1(Writers.scala:134)
	at upickle.implicits.Writers$$anon$18.$anonfun$write0$1$adapted(Writers.scala:130)
```

Am I missing something?

PR #436, opened 2023-01-31
Ensure that ujson.Value behaves properly as a direct target of reading/writing

lihaoyi:
The basic issue is that the `StringVisitor`/`StringReader` used for processing map keys when building JSON ASTs (https://github.com/com-lihaoyi/upickle/blob/d0b618e8b393089ca0e2bbea6648732a62227d9a/ujson/src/ujson/AstTransformer.scala#L30) was not sufficiently forgiving in accepting non-string types, which can be given to it when serializing a map with non-string keys.

This "loosening" of the input handling is in line with the general trend throughout uPickle, similar to how in 2.0.0 we allowed  https://github.com/com-lihaoyi/upickle/issues/385 we allowed numeric Scala types to accept JSON strings as input

This PR adds it to the test suite conversion matrix, so this code path is additionally exercised for every value in our test suite, and fixed another bug uncovered that crashed when directly writing `Infinity` and `-Infinity` to `ujson.Value`

Fixes https://github.com/com-lihaoyi/upickle/issues/433


lefou (commented):
Looks reasonable to me. The binary breakage is ok here, but it's interesting, that it only breaks for Scala.js.

lefou:
> Looks reasonable to me. The binary breakage is ok here, but it's interesting, that it only breaks for Scala.js.

Oh, it runs without `--keep-going`, so it probably also fails for all other platforms.

lihaoyi:
We're going to release 3.0.0 next anyway right? which would mean binary compat breakage doesn't matter?

lefou:
> We're going to release 3.0.0 next anyway right? which would mean binary compat breakage doesn't matter?

Exactly. The whole point about the `3.0.0-M1` pre-release was to get some binary breaking things in.

PR #437, opened 2023-01-31
CI: Continue failing build to see more issues

lefou:


.github/workflows/actions.yml
```diff
@@ -28,8 +28,7 @@ jobs:
           java-version: ${{ matrix.java }}
       - name: Run JS tests
         run: |
-          ./mill -i __.resolvedIvyDeps
-          ./mill -i -j 0 all __.js.__.compile __.js.__.publishArtifacts __.js.__.test
+          ./mill -i -j 0 -k __.js.__.compile + __.js.__.publishArtifacts + __.js.__.test
```
lefou, line 31: Mill 0.10.10 is mitigating most of the known coursier issues, so there is no need to run `__.resolvedIvyDeps` in non-parallel mode. 🤞‍

.github/workflows/actions.yml
```diff
@@ -83,7 +80,7 @@ jobs:
           distribution: 'temurin'
           java-version: 8
       - name: Check Binary Compatibility
-        run: ./mill -i __.mimaReportBinaryIssues
+        run: ./mill -i -k __.mimaReportBinaryIssues
```
lefou, line 83: We want to see all binary compatibility issues.

PR #438, opened 2023-01-31
Replaced the 'µ' with 'u' in "µPickle"

lefou:


nafg:
Why?

lefou:
> Why?

It was the only occurrence of the word "µPickle". In all other places this librarry is called "uPickle".

Issue #389, opened 2022-04-30
Scala 3 serialization macros are very inefficient compared to Scala 2


lihaoyi:
Compared Scala 2 (https://github.com/com-lihaoyi/upickle/tree/master/implicits/src-2/upickle/implicits) and Scala 3 (https://github.com/com-lihaoyi/upickle/tree/master/implicits/src-3/upickle/implicits) implementations. 

The Scala 2 implementation does not contain any runtime logic at all, and generates a single class at compile time for each case class in question. During reading, temporary values are stored unboxed in typed case class fields, accessed as fields, and put directly into class constructor. During writing, the case class fields are read and put directly into the visitor. That is all that happens at runtime.

The Scala 3 implementation does a bunch of runtime reflection to get fields, puts values boxed in temporary lists and `Map[String, Any]`s, does dictionary lookups, and all sorts of other things like that. While it passes tests, it is likely an order of magnitude slower than the Scala 2 implementation, and really should be re-written to be in the same style: do all reflection at compile-time, generating lean and efficient code to do minimal work at runtime

jodersky:
I am happy to take a look at it. It would be a good opportunity to also check #386.

lihaoyi:
Thanks @jodersky! I was planning on releasing 2.0.0 since the other PRs have merged, but if you're going to take a look at this I can hold off a week to see if we can include your work. The proposed changes will affect binary compatibility for Scala 3 if I'm not mistake

jodersky:
> the proposed changes will affect binary compatibility for Scala 3 if I'm not mistake

are you talking about binary compatibility of 2.0 or macros? In case you mean the latter, I don't think that anything would change. The macros are a black box, right? They will generate instances of the ReadWriters, but beyond that they are totally opaque to user programs. What I'm saying is that while I appreciate you willing to hold off on 2.0 for a little, I don't think you have to :)  We could release fixes to macros in a 2.1 or other patch release.

lihaoyi:
@jodersky I'm imagining a case where someone compiles against scala3/upickle2 with the current macros+runtimesupport, we release a new scala3/upickle2+ without the runtime support, and then someone using their their earlier-compiled macro serializers against the newer version of upickle hits NoSuchMethod errors. The earlier-compiled code would still have callsites to the runtime support code which gets removed in upickle 2+

[...]

Issue #388, opened 2022-04-23
Scala 3 recursive structure go into infinite loop


liewhite:
Hello, deriving upickle's ReadWriter for recursive structures in scala 3  seems run into infinite loop...
```scala
case class Recur(recur: Option[Recur]) derives ReadWriter
println(write(Recur(None))) // hang on here
```
And with no compile error
Maybe we should use lazy vals somewhere?

aaronp:
I've just come here to report this as well. I'm creating a swagger generator for parts of Li's stack (cask for server, requests for client), and the 'petstore' test case has a recursive data structure like this - exposing this issue.

I'd love to help if I can - pairing on this, whatever. Otherwise +1 for the issue

lefou:
Closed by #440

PR #440, opened 2023-02-01
Speeding up Scala 3 Writing macros

lihaoyi:
Fixes https://github.com/com-lihaoyi/upickle/issues/389 and https://github.com/com-lihaoyi/upickle/issues/388

This PR moves a bunch of logic from runtime to compile time, and optimizes whatever runtime logic remains. It also consolidates the logic with the Scala 2 logic as much as possible, e.g. re-using `writeSnippet` to manage writes and  `CaseObjectContext`/`HugeCaseObjectContext` to manage validation and error reporting during a read

Running some ad-hoc benchmarks on my laptop, `./mill bench.jvm.run`, it gives a significant speedup on reads and writes, and brings it close to the Scala 2.13 numbers (higher is better):

| Benchmark (5000ms)                  | Scala 3 Before | Scala 3 After | Scala 2 |      
|-------------------------------------|----------------|---------------|---------|      
| upickleDefault Read                 |      637       |     1065      |   1403  |      
| upickleDefault Write                |      839       |     1452      |   1549  |      
| upickleDefaultByteArray Read        |      582       |     1172      |   1126  |      
| upickleDefaultByteArray Write       |      847       |     1218      |   1277  |      
| upickleDefaultBinary Read           |      925       |     3853      |   3844  |      
| upickleDefaultBinary Write          |      1252      |     3117      |   3666  |      
| upickleDefaultCached Read           |      620       |     1300      |   1412  |      
| upickleDefaultCached Write          |      829       |     1555      |   1588  |      
| upickleDefaultByteArrayCached Read  |      575       |     1182      |   1095  |      
| upickleDefaultByteArrayCached Write |      838       |     1223      |   1297  |      
| upickleDefaultBinaryCached Read     |      928       |     3825      |   3885  |      
| upickleDefaultBinaryCached Write    |      1266      |     2907      |   3674  |      

Note that the generated code, especially for reads, is still not as optimized as the Scala 2 versions:

1. I couldn't figure out how to generate an anonymous class with typed fields in Scala 3, so I'm putting things in an `Array[Any]`
2. I couldn't figure out how to generate `match` statements, so I generated a `Map[K, V]` and look it up at runtime.

Fixing this issues and moving the reading logic into the macro should also be possible, but can happen in a separate PR

All existing tests pass. Added a regression test for the recursive Scala 3 scenario that hangs on master. Also moved a bunch of `AdvancedTests` into the shared folder now that they work


core/src/upickle/core/Types.scala
```diff
@@ -277,16 +277,16 @@ trait Types{ types =>
...
-    protected def writeSnippet[R, V](objectAttributeKeyWriteMap: CharSequence => CharSequence,
-                                     ctx: _root_.upickle.core.ObjVisitor[_, R],
-                                     mappedArgsI: String,
-                                     w: Writer[V],
-                                     value: V) = {
+    def writeSnippet[R, V](objectAttributeKeyWriteMap: CharSequence => CharSequence,
+                           ctx: _root_.upickle.core.ObjVisitor[_, R],
+                           mappedArgsI: String,
+                           w: Any,
+                           value: Any) = {
```
lihaoyi, line 284: Had to make these un-typed because I couldn't figure out how to make the Scala 3 macro quotes/splices play nicely with the path dependent `prefix.Writer` type. A bit gross but shouldn't affect perf or correctness

core/src/upickle/core/Types.scala
```diff
@@ -277,16 +277,16 @@ trait Types{ types =>
         ctx.visitEnd(-1)
       }
     }
-    protected def writeSnippet[R, V](objectAttributeKeyWriteMap: CharSequence => CharSequence,
-                                     ctx: _root_.upickle.core.ObjVisitor[_, R],
-                                     mappedArgsI: String,
-                                     w: Writer[V],
-                                     value: V) = {
+    def writeSnippet[R, V](objectAttributeKeyWriteMap: CharSequence => CharSequence,
```
lihaoyi, line 280: I couldn't figure out how to subclass the path dependent `prefix.CaseW` type in the macro, so I had to instantiate it in the inline part first and then call into the macro after. That means the call to `writeSnippet` is no longer from the subclass, and it thus can no longer be protected

implicits/src-3/upickle/implicits/Readers.scala
```diff
@@ -1,9 +1,117 @@
 package upickle.implicits
```
lihaoyi, line 1: This file is mostly just consolidated from `CaseClassReader.scala`; proper rewrite would be a separate PR

implicits/src-3/upickle/implicits/Readers.scala
```diff
@@ -1,9 +1,110 @@
...
+        macros.checkErrorMissingKeysCount[T]()
+      )
+
+      if macros.isMemberOfSealedHierarchy[T] then annotate(reader, macros.fullClassName[T])
+      else reader
+
+    case m: Mirror.SumOf[T] =>
+      inline compiletime.erasedValue[T] match {
+        case _: scala.reflect.Enum =>
+          val valueOf = macros.enumValueOf[T]
```
bishabosha, line 92: this won't work for ADT enums, the solution here is to assert all cases are singleton:
```scala
summonAll[Tuple.Map[m.MirroredElemTypes, [S] =>> S <:< Singleton]] // assert all singleton
```

lihaoyi:
I'll leave this over the weekend, and if there aren't any more comments I'll merge it then. All the existing/new/newly-enabled tests pass, and the benchmarks demonstrate the non-functional improvements, so I can't have broken anything too badly

The issue noticed by @bishabosha is a limitation in existing code, which is why it was not picked up by any tests (since it never worked). Fixing that can come in a follow up

lefou (approved):
This looks good to me, although I'm myself not experienced in Scala 3 Macros. But tests pass and benchmark looks promising.

After we merge this, I'd like to cut a next milestone release `3.0.0-M2` to get this changed version tested in as many places as possible.

lefou:
There seem to be issues with this change. E.g. in Ammonite, the integration tests don't succeed. 

* https://github.com/com-lihaoyi/Ammonite/pull/1320

PR #441, opened 2023-02-06
Update MiMa baseline to 3.0.0-M2

lefou:


lefou:
There seem to be issues with this change. E.g. in Ammonite, the integration tests don't succeed. 

* https://github.com/com-lihaoyi/Ammonite/pull/1320

PR #445, opened 2023-02-27
Normalize handling of Scala 3 enums, remove allocations from reader

### LINKED ISSUE #356: Writing Scala 3 enums is broken

lihaoyi:
Fixes #356

The Scala 3 code for handling enums was a parallel code path from that handling sealed hierarchies: going through separate `EnumReader`/`EnumWriter`s, raising separate error messages, and not supporting common use cases. This PR makes Scala 3 enums use the same sealed-trait-handling logic as far as possible, allowing them to support things like parametrized enums and parametrized enum cases. This also reduces the complexity of the code by ensuring both `enum`s and `sealed trait`s go through a common code path, and should help avoid accidental divergence in future

One notable point of divergence is the choice of type tag. Enums use the short name of the type `Foo`, whereas sealed traits use the fully qualified name `pkg.name.Foo`. For now, I leave both behaviors in place, though ideally we should unify them at some point

I also took the opportunity to tweak the Scala 3 `Reader.scala` macros, to move more logic to compile time to better mirror that behavior in Scala 2. This replaces a runtime `Map` in each case class reader with a macro-generated `match` statement, and reduces the number of runtime `Arrays` we need to allocate (though does not quite eliminate all of them, we still have two arrays for `params` and `visitors`)

All existing tests pass, and added new tests to verify the newly-working functionality. 

I'll leave this up for a few days in case anyone wants to review it

### DIFF
--- a/core/src/upickle/core/Types.scala
+++ b/core/src/upickle/core/Types.scala
@@ -338,24 +338,19 @@ trait Types{ types =>
     override def expectedMsg = taggedExpectedMsg
     override def visitArray(length: Int, index: Int) = taggedArrayContext(this, index)
     override def visitObject(length: Int, jsonableKeys: Boolean, index: Int) = taggedObjectContext(this, index)
-    override def visitString(s: CharSequence, index: Int) = findReader(s.toString).visitString(s, index)
+    override def visitString(s: CharSequence, index: Int) = {
+      findReader(s.toString) match {
+        case null => throw new AbortException("invalid tag for tagged object: " + s, index, -1, -1, null)
+        case reader => reader.visitString(s, index)
+      }
+    }
   }
   object TaggedReader{
     class Leaf[T](tag: String, r: Reader[T]) extends TaggedReader[T]{
-      def findReader(s: String) = {
-        println("LHY findReader s " + s)
-        println("LHY findReader tag " + tag)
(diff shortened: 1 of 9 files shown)


lefou (commented):
I glanced over these changes and they look good to me. This is a great PR. Thank you to all who helped! Yet, I don't fully understand all details and lack the time to do so in the near future, so I'd welcome additional reviews.

PR #447, opened 2023-03-02
Be more robust about checking whether objects are modules in Scala3

lihaoyi:
`.isSingleton` doesn't work reliable in the presence of `derives`, neither does `valueOf[T]`. This PR re-implements our own versions that work properly


lefou (approved):
Looks good to me.

Issue #446, opened 2023-02-27
DoS vulnerability with ujson


lolgab:
`ujson` is vulnerable to a DOS Attack where the attacker tries to read a json object where every key has the same hashcode. This vulnerability originates in the scala `LinkedHashMap` implementation.
As [suggested](https://github.com/com-lihaoyi/upickle/issues/416#issuecomment-1446005154) by @plokhotnyuk, a solution would be to use, instead of [scala.collection.mutable.LinkedHashMap](https://www.scala-lang.org/api/2.13.10/scala/collection/mutable/LinkedHashMap.html), a Scala wrapper around [java.util.LinkedHashMap](https://docs.oracle.com/javase/8/docs/api/java/util/LinkedHashMap.html).

<details>
<summary>Click to expand a self-contained reproduction for the issue, copy pasted from the jsoniter-scala benchmark</summary>

```scala
object Main {
  def zeroHashCodeStrings: Iterator[String] = {
    def charAndHash(h: Int): Iterator[(Char, Int)] = ('!' to '~').iterator.map(ch => (ch, (h + ch) * 31))

    for {
      (ch0, h0) <- charAndHash(0)
      (ch1, h1) <- charAndHash(h0)
      (ch2, h2) <- charAndHash(h1) if ((h2 + 32) * 923521 ^ (h2 + 127) * 923521) < 0
      (ch3, h3) <- charAndHash(h2) if ((h3 + 32) * 29791 ^ (h3 + 127) * 29791) < 0
      (ch4, h4) <- charAndHash(h3) if ((h4 + 32) * 961 ^ (h4 + 127) * 961) < 0
      (ch5, h5) <- charAndHash(h4) if ((h5 + 32) * 31 ^ (h5 + 127) * 31) < 0
      (ch6, h6) <- charAndHash(h5) if (h6 + 32 ^ h6 + 127) < 0
      (ch7, _) <- charAndHash(h6) if h6 + ch7 == 0
    } yield new String(Array(ch0, ch1, ch2, ch3, ch4, ch5, ch6, ch7))
  }

  val jsonString =
    zeroHashCodeStrings
      .map(s => ujson.write(s))
      .take(1000000)
      .mkString("{", s":null,", ":null}")

  def main(args: Array[String]): Unit = {
    ujson.read(jsonString)
  }
}
```
</details>

PR #449, opened 2023-03-07
Fix DoS vunerability with ujson.Obj and upack.Obj

lolgab:
Fixes: #446

## Replace `scala.collection.mutable.LinkedHashMap` with `upickle.core.LinkedHashMap` in `ujson.Obj` and `upack.Obj`

`scala.collection.mutable.LinkedHashMap` is not secure because of https://github.com/com-lihaoyi/upickle/issues/446
An attacker can create a sequence of keys that return the same `hashCode`.
This PR replaces it with a bespoke wrapper of `java.util.LinkedHashMap` which extends `scala.collection.mutable.Map` since `java.util.LinkedHashMap` is safe against this kind of hash-collision attacks.


core/src/upickle/core/internal/LinkedHashMap.scala
```diff
@@ -0,0 +1,54 @@
...
+import upickle.core.compat.Factory
+
+import scala.collection.mutable
+import java.{util => ju}
+
+class LinkedHashMap[K, V] private (underlying: ju.LinkedHashMap[K, V])
+    extends mutable.Map[K, V]
+    with mutable.Builder[(K, V), LinkedHashMap[K, V]]
+    with LinkedHashMapCompat[K, V]
+    {
```
lefou, line 12: Is this a copy of somewhere else or does it need a thorough review?

Some ScalaDoc would be nice. It should at least tell that it is a mutable collection.

lefou (commented):
You change the visibility of various case classes. Is this also motivated by improving binary compatibility story, or only to better control the construction process? If the former, there is more to do, at least for Scala 2, namely hiding the `copy` and the `unapply` methods.

lolgab:
I made the `Obj` constructor private and removed `case` so there is no generated apply method. I want to keep a private constructor to wrap a `Map` so the internal implementation can take advantage of it, while make sure users can't wrap a map that doesn't keep ordering without copying it to a LinkedHashMap.
I privatized the other constructors for consistency since apply is the documented way to instantiate the object  and having some that can do `new Thing` and some that can't seemed inconsistent.
It's not my goal to improve the binary compatibility.

core/src/upickle/core/internal/LinkedHashMap.scala
lolgab: I wrote it from scratch. So needs a proper review.
The implementation is internal to this repository, also it extends `scala.collection.mutable.Map`.
Nevertheless I can add the Scaladoc.

lihaoyi:
Rather than privatizing all the constructors, could we just publicize `upickle.internal.LinkedHashMap` as `upickle.JsonHashMap` and make it the official type signature `case class Obj(value: JsonHashMap)`? That would greatly reduce the churn in the API, keep everything as "simple" case classes, while still leaving us the option to update the `JsonHashMap` implementation in future (e.g. to specialize it for strings for perf)

lolgab:
@lihaoyi We also have upack.. So maybe it can be an `upickle.core.internal.LinkedHashMap` and two public implementations, one with `[String, ujson.Value]` and one with `[upack.Msg, upack.Msg]`?
But `StringHashMap` as a name doesn't communicate the fact that the value is of type `ujson.Value`.
Maybe something like `ujson.ObjMap` can be better?
Or making `ujson.Obj` itself extend `internal.LinkedHashMap[String, Value]`? We could provide `.value` as just a method like `def value = this`, but this would add a lot of bloat from `scala.collection.mutable.Map` to `Obj` API, which maybe you don't want to be exposed to if you don't call `.value`.

lihaoyi:
I don't think we need to communicate it's only used in ujson.Value. As you said, it's used in upack as well. It can just be a generic "mutable string map preserving input ordering" class.

StringMap or ObjMap both sound good to me

I don't think we should overhaup the ujson.Obj API in this PR. Not because it doesn't need work, just that that work is pretty orthogonal to fixing this particular performance issue. Regardless, if we wanted to add more helpers on ujson.Obj/ujson.Value it should be possible later without bincompat breakage

lolgab:
The problem is that `upack` doesn't have `String` as the key type but `upack.Msg`

lefou:
If this map is a pragmatic mostly internal structure, lets find a pragmatic name paired with some ScalaDoc. What about `UjsonLinkedHashMap`, it transports it's properties (hash map, linked) but also that it's tied to `j{json,pack,pickle}`. The curious user can read the scaladoc to read, why it is needed and whether it is safe to be used in other places.

lihaoyi:
sounds good. I think just `upickle.core.LinkedHashMap` would be fine as well, so not specific to either ujson or upack

lolgab:
@lihaoyi @lefou Ready for another review round

lihaoyi (approved):
Looks good to me i think

core/src-2.12/upickle/core/compat/LinkedHashMapCompat.scala
```diff
@@ -0,0 +1,27 @@
...
+        private val map = LinkedHashMap[K, V]()
+
+        def +=(elem: (K, V)): this.type = {
+          map.addOne(elem)
+          this
+        }
+        def clear(): Unit = map.clear()
+        def result(): LinkedHashMap[K, V] = map
+      }
+      def apply(it: Nothing) = ???
```
lefou, line 25: ???

ujson/src/ujson/Value.scala
```diff
@@ -1,13 +1,9 @@
 package ujson
 
-
-
-import upickle.core.Util
-import upickle.core.{ObjArrVisitor, Visitor}
-
+import upickle.core._
```
lefou, line 3: Can we keep explicit imports?

upickle/test/src/upickle/AdvancedTests.scala
```diff
@@ -357,19 +357,27 @@ object AdvancedTests extends TestSuite {
...
-        // which is not robust against collisions
-        //        ujson.read(jsonString)
-
-        // Make sure this can pass in Scala 3. We do not use `ujson.Value` in this code path at all,
-        // so it should finish in a reasonable amount of time and without issue
-        upickle.default.read[Foo416](jsonString)
+        test("issue-416") {
+          upickle.default.read[Foo416](jsonString(1000000))
+          ()
+        }
```
lefou, line 369: In what way will this test catch the issue?

upickle/test/src/upickle/AdvancedTests.scala
```diff
@@ -357,19 +357,27 @@ object AdvancedTests extends TestSuite {
...
+          sys.props("java.vm.name") match {
+            case "Scala.js" | "Scala Native" =>
+            // The fix assumes a hash collision safe java.util.LinkedHashMap
+            // implementation. When/if other platforms will have such characteristics
+            // ujson doesn't need to change.
+            case _ =>
+              ujson.read(jsonString(100000))
+          }
+          ()
+        }
```
lefou, line 380: In what way will this test catch the issue?

ujson/src/ujson/Value.scala
lolgab: `import upickle.core.{LinkedHashMap, ObjArrVisitor, Util, Visitor}` seems a bit too long, no?

core/src-2.12/upickle/core/compat/LinkedHashMapCompat.scala
lolgab: We define `type Factory[-A, +C] = CanBuildFrom[Nothing, A, C]` exactly because we don't need this second method.
Since there is no value of type `Nothing` in Scala this is safe to be unimplemented since it's impossible to call.

lefou: Ok, Thanks. 👍‍

ujson/src/ujson/Value.scala
lefou: Not to me. I love when I can read source code without an IDE, which is not safely possible when there are multiple wildcard imports.

upickle/test/src/upickle/AdvancedTests.scala
lolgab: Using the previous implementation this was not completing. I'm not sure that if I waited more, it would actually complete, but since it's a performance problem I don't know how to test it. Putting time constraints leads to flaky tests, so I would avoid them.

upickle/test/src/upickle/AdvancedTests.scala
lolgab: Using the previous implementation this was not completing. I'm not sure that if I waited more, it would actually complete, but since it's a performance problem I don't know how to test it. Putting time constraints leads to flaky tests, so I would avoid them.

upickle/test/src/upickle/AdvancedTests.scala
lefou: Hm, I think a Future/Await with a reasonable too long time, which may fail once in a month, is better than an one half day running but succeeding test in CI, which nobody recognizes as failure.

lefou: It will also speed up the failure result.

ujson/src/ujson/Value.scala
lefou: This is even more true, when a type like `LinkedHashMap` also exists in the standard lib. A quick look at the imports shows where it really comes from.

lefou (approved):
Looks good to me. This was lots of work to be on the safe side. Nice job.

Can you update the PR description. I'd like to use it as commit message, too.

lolgab:
Thank you very much @lefou!
I updated the PR description.

PR #454, opened 2023-03-10
Attempt to properly generate enum case serializers without global implicits

lihaoyi:
Follow on from https://github.com/com-lihaoyi/upickle/pull/453. Doesn't work yet


lihaoyi:
Superseded by https://github.com/com-lihaoyi/upickle/pull/453

PR #458, opened 2023-03-10
circe compatibility with scala 3.1.1

Quafadas:
https://github.com/com-lihaoyi/upickle/pull/452#discussion_r1132777964


lolgab:
We now updated Scala 3 to `3.2.2` so this is not needed anymore

Issue #456, opened 2023-03-10
`ujson.Bool.unapply` generates `match may not be exhaustive` warning


lolgab:
The following code:
```scala
ujson.Bool(true) match {
  case ujson.Bool(v) =>
}
```
generates the following warning:
```scala
ujson.Bool(true) match {
^^^^^^^^^^^^^^^^
match may not be exhaustive.

It would fail on pattern case: False, True
```

The problem is in the `unapply` signature:
```scala
def unapply(bool: Bool): Option[Boolean] = Some(bool.value)
```

should instead be:
```scala
def unapply(bool: Bool): Some[Boolean] = Some(bool.value)
```

lolgab:
To be able to test this we need to enable `-Xfatal-warnings` so I opened #457 to do that.

PR #461, opened 2023-03-11
Fix Bool pattern matching to be exhaustive

lolgab:
Fixes https://github.com/com-lihaoyi/upickle/issues/456

### DIFF
--- a/upack/test/src/upack/BoolTests.scala
+++ b/upack/test/src/upack/BoolTests.scala
@@ -1,0 +1,32 @@
+package upack
+
+import utest._
+
+object BoolTests extends TestSuite {
+  def tests = Tests{
+    test("upack.Bool apply") {
+      upack.Bool(true) ==> upack.True
+      upack.Bool(false) ==> upack.False
+    }
+
+    test("upack.Bool.value") {
+      upack.Bool(true).value ==> true
+      upack.Bool(false).value ==> false
+    }
+    test("upack.Bool unapply") {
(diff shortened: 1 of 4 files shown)


upack/test/src/upack/BoolTests.scala
```diff
@@ -0,0 +1,32 @@
...
+          case upack.Bool(value)
+            if value == bool
+            && jsb.value == value =>
+        }
+      }
+    }
+    test("upack.Bool pattern matching should be exhaustive") {
+      upack.Bool(true) match {
+        case upack.Bool(value) =>
+      }
```
lefou, line 29: This test assumes we're using `-Xfatal-warnings`, right? If not, I don't see how this test is supposed to detect non-exhaustiveness.

lolgab: Yes, exactly. I added it in the previous PR exactly for this test. I don't know if there is a better way to test this.

lefou: That means, if the settings get changed later, we have no indicator that this test is insufficient. The bare minimum is a comment in code or in the test name. A bit better would be to detect the flags (e.g. by generating some build info code). I know utest has `compileError`; maybe it also fails a warning under `-Xfatal-warnings`, in which case we could guard some code that emits a warning with `compileError` to detect when this setting gets removed in the future.

lihaoyi:
Gonna merge this to get it in for the 3.0.0 release

PR #473, opened 2023-03-20
Release uPickle 3.1.0

lihaoyi:
Targeting a release end-April. Since none of the fixes are showstoppers (the bugs have workarounds, and perf improvements are always just a nice-to-have) there isn't any real urgency to push them out, and we can wait a bit to see if anyone else has changes they want to get in


lihaoyi:
Since nothing much has merged since this opened, I'm inclined to release this sooner rather than later. Will merge and release 3.1.0 tomorrow

PR #579, opened 2024-05-14
Support customizing `$type` tag with `key` annotation

### LINKED ISSUE #578: Allow `$type` tag to be replaced by some other string (500USD Bounty)
Given a `"$type": "foo.Bar"` sealed trait discriminator, [Custom Keys](https://com-lihaoyi.github.io/upickle/#CustomKeys)'s `@key` annotation allows you to replace the right hand side `"foo.Bar"` of a particular `case class` by annotating the `case class`, but it does not allow you to replace to left hand side `"$type"` string with something else. 

This ticket is to implement the ability for a user to replace `$type"` with some other string, by annotating the `sealed trait` itself with `@key("myType")` annotation that would replace `"$type"` with `"myType"` for all sub-classes of the `sealed trait`.

To incentivize contribution, I'm putting a 500USD bounty on resolving this ticket. This is payable via bank transfer, and at my discretion in case of ambiguity.

mrdziuban:
Closes #578 

Adds support for customizing the key used a discriminator for sealed hierarchies. The default is still `$type`, but can be customized by annotating the `sealed trait` with `@key("customKey")`.

The code changes necessary to support this in `Writer` were:
- Adding a new `tagKey` parameter to `TaggedWriter.Leaf`
- Adding a new `TaggedWriter#findWriterWithKey` method that includes the `tagKey` in its return value
- Updating `TaggedWriter#write0` to call `findWriterWithKey` and pass the `tagKey` through to `taggedWrite`
- Updating `AttributeTagged#taggedWrite` to use the given `tagKey` in place of the old default key

The code changes necessary to support this in `Reader` were:
- Adding a new `tagKey` method to `TaggedReader`, which defaults to the old default key
- Adding a `tagKey` parameter to `TaggedReader.Leaf`, `TaggedReader.Node`, `TaggedReadWriter.Leaf`, and `TaggedReadWriter.Node` to override the default value in `TaggedReader`
- Updating `AttributeTagged#taggedObjectContext` to use the given `TaggedReader`'s `tagKey` in place of the old default key

I tried instead to pass the `tagKey` through to `findReader`, but `findReader` is called in a number of contexts where it wasn't possible to get the `tagKey` value, for example in `TaggedReader#visitString` and the `visitValue` method within the `ObjVisitor` returned by `AttributeTagged#taggedObjectContext`.

The one limitation of the customization is when a member of a sealed hierarchy inherits from multiple parents that would result in different discriminators, e.g. if two parents have `@key` annotations with different arguments, or if one parent has a `@key` annotation and another doesn't. In these cases, the macros will fail with an error telling the user what went wrong and suggesting ways to fix it:

```scala
@key("customKey1")
sealed trait MultiKeyedADT1
@key("customKey2")
sealed trait MultiKeyedADT2
case object MultiKeyedObj extends MultiKeyedADT1 with MultiKeyedADT2 {
  implicit val rw: ReadWriter[MultiKeyedObj.type] = macroRW
}
-- Error: ----------------------------------------------------------------------
 6 |  implicit val rw: ReadWriter[MultiKeyedObj.type] = macroRW
   |                                                    ^^^^^^^
   |Type MultiKeyedObj.type inherits from multiple parent types with different discriminator keys:
   |
   |  parents: MultiKeyedADT1, MultiKeyedADT2
   |  keys: customKey1, customKey2
   |
   |To resolve this, either remove the `@key` annotations from all parents of the type,
   |or make sure all the parents pass the same value to `@key`
```

### DIFF
--- a/upickle/test/src/upickle/ClassDefs.scala
+++ b/upickle/test/src/upickle/ClassDefs.scala
@@ -55,7 +55,7 @@ object ADTs {
 object Hierarchy {
   sealed trait A
   object A{
-    implicit def rw: upickle.default.ReadWriter[A] = RW.merge(B.rw, C.rw)
+    implicit def rw: upickle.default.ReadWriter[A] = RW.merge(core.Annotator.defaultTagKey, B.rw, C.rw)
   }
   case class B(i: Int) extends A
   object B{
@@ -68,6 +68,7 @@ object Hierarchy {
 
   object Z{
     implicit def rw: upickle.default.ReadWriter[Z] = RW.merge(
+      core.Annotator.defaultTagKey,
       implicitly[upickle.default.ReadWriter[AnZ.type]]
     )
   }
@@ -79,7 +80,7 @@ object Hierarchy {
 object DeepHierarchy {
   sealed abstract class A
   object A{
-    implicit def rw: RW[A] = RW.merge(B.rw, C.rw)
+    implicit def rw: RW[A] = RW.merge(core.Annotator.defaultTagKey, B.rw, C.rw)
   }
   case class B(i: Int) extends A
 
@@ -90,7 +91,7 @@ object DeepHierarchy {
   sealed trait C extends A
 
   object C{
-    implicit def rw: RW[C] = RW.merge(D.rw, E.rw, F.rw)
+    implicit def rw: RW[C] = RW.merge(core.Annotator.defaultTagKey, D.rw, E.rw, F.rw)
   }
   case class D(s: String) extends C
 
@@ -106,7 +107,7 @@ object DeepHierarchy {
   sealed trait Q //new line
 
   object Q{
-    implicit def rw: RW[Q] = RW.merge(AnQ.rw)
+    implicit def rw: RW[Q] = RW.merge(core.Annotator.defaultTagKey, AnQ.rw)
   }
   case class AnQ(i: Int) extends Q //new line
 
@@ -125,6 +126,7 @@ object Singletons{
 
   object AA{
     implicit def rw: RW[AA] = RW.merge(
+      core.Annotator.defaultTagKey,
       default.macroRW[BB.type],
       default.macroRW[CC.type]
    )
@@ -154,7 +156,7 @@ object Generic{
 object Recursive{
   sealed trait LL
   object LL{
-    implicit def rw: RW[LL] = RW.merge(default.macroRW[End.type], default.macroRW[Node])
+    implicit def rw: RW[LL] = RW.merge(core.Annotator.defaultTagKey, default.macroRW[End.type], default.macroRW[Node])
   }
   case object End  extends LL
   case class Node(c: Int, next: LL) extends LL
@@ -167,7 +169,7 @@ object Recursive{
   }
   sealed trait SingleTree
   object SingleTree{
-    implicit def rw: RW[SingleTree] = RW.merge(default.macroRW[SingleNode])
+    implicit def rw: RW[SingleTree] = RW.merge(core.Annotator.defaultTagKey, default.macroRW[SingleNode])
   }
   case class SingleNode(value: Int, children: List[SingleTree]) extends SingleTree
   object SingleNode{
@@ -177,7 +179,7 @@ object Recursive{
 object Annotated {
   sealed trait A
   object A{
-    implicit def rw: RW[A] = RW.merge(default.macroRW[B], default.macroRW[C])
+    implicit def rw: RW[A] = RW.merge(core.Annotator.defaultTagKey, default.macroRW[B], default.macroRW[C])
   }
   @key("0") case class B(@key("omg") i: Int) extends A
   object B{
@@ -312,7 +314,7 @@ object Exponential{
 object GenericADTs{
   sealed trait Small[A]
   object Small{
-    implicit def rw[A: R: W]: RW[Small[A]] = RW.merge(Small1.rw[A])
+    implicit def rw[A: R: W]: RW[Small[A]] = RW.merge(core.Annotator.defaultTagKey, Small1.rw[A])
   }
   case class Small1[A](key: A) extends Small[A]
   object Small1{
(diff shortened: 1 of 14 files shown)


lihaoyi:
@mrdziuban I think the approach looks great.

> Determine how to handle cases where a type inherits from multiple sealed traits

In the current codebase, I suppose it doesn't matter, because the parent `sealed trait` that a `case class` inherits from does not change the serialization format. With this change, that is no longer true.

I think a reasonably first pass would be to just fail loudly if a `case class` inherits from multiple `sealed trait`s with different `@key`s configured (or some configured and others not configured). The "different parent `sealed trait`s have different discriminator fields" use case is probably obscure enough most people won't notice.

Another alternative would be to make the `case class` have multiple duplicate discriminator fields if its parents have inconsistent `@keys`, e.g. `{"$type": "foo.Bar", "$type2": "foo.Baz", "data", "blah", ...}`. But I think that can be left to a follow PR, and erroring out now gives us the flexibility to adjust the design in future if we learn more

 > Consider if there's a more ergonomic way to get tagKey in [taggedObjectContext](https://github.com/com-lihaoyi/upickle/blob/b472fc9994a2cbdd2f93027f33a2d0366d7a8968/upickle/src/upickle/Api.scala#L320)

Could you describe what you think is un-ergonomic about the current approach? That would help me give a proper answer

mrdziuban:
> fail loudly if a case class inherits from multiple sealed traits with different @keys configured (or some configured and others not configured)

Good idea, updated to do so.

> Could you describe what you think is un-ergonomic about the current approach?

I wasn't sure if adding `def tagKey` to `TaggedReader` was the best approach, since it proliferates to everywhere that constructs one, like `Reader.merge` and `Writer.merge`. That said, I tried some other approaches, like passing the `tagKey` to `findReader`, but ran into limitations with all of them where I didn't have the value available where `findReader` was called.

mrdziuban:
Re: binary compatibility, I've made progress on it but am still having trouble with 5 problems:

```
 * abstract method annotate(upickle.core.Types#Reader,java.lang.String,java.lang.String)upickle.core.Types#TaggedReader in interface upickle.core.Annotator is present only in current version
   filter with: ProblemFilter.exclude[ReversedMissingMethodProblem]("upickle.core.Annotator.annotate")

 * abstract method annotate(upickle.core.Types#ObjectWriter,java.lang.String,java.lang.String,upickle.core.Annotator#Checker)upickle.core.Types#TaggedWriter in interface upickle.core.Annotator is present only in current version
   filter with: ProblemFilter.exclude[ReversedMissingMethodProblem]("upickle.core.Annotator.annotate")

 * abstract method taggedWrite(upickle.core.Types#ObjectWriter,java.lang.String,java.lang.String,upickle.core.Visitor,java.lang.Object)java.lang.Object in interface upickle.core.Types is present only in current version
   filter with: ProblemFilter.exclude[ReversedMissingMethodProblem]("upickle.core.Types.taggedWrite")

 * private[..] abstract method tagKey()java.lang.String in interface upickle.core.Types#TaggedReader is present only in current version
   filter with: ProblemFilter.exclude[ReversedMissingMethodProblem]("upickle.core.Types#TaggedReader.tagKey")
   
 * abstract method findWriterWithKey(java.lang.Object)scala.Tuple3 in interface upickle.core.Types#TaggedWriter is present only in current version
   filter with: ProblemFilter.exclude[ReversedMissingMethodProblem]("upickle.core.Types#TaggedWriter.findWriterWithKey")
```

I can fix the first two by implementing the `annotate` methods in `Types.scala` to construct the `TaggedReader.Leaf` and `TaggedWriter.Leaf` instances like they do in the `LegacyApi` and `AttributeTagged` in `Api.scala`, but I'm not sure if there's any drawback to this and/or if they need to stay abstract in `Types.scala`. These are the methods I'm talking about: https://github.com/com-lihaoyi/upickle/blob/d050f729197a96db7bf8fcde9a6d8645e70126c4/upickle/core/src/upickle/core/Types.scala#L287-L288

I'm stumped on the others. Do you have any idea how to handle them?

lihaoyi:
The `abstracy method present only in current version` errors generally can be resolved by giving the method some default implementation. e.g. 

- if the new abstract method has an additional argument, make it instead a concrete method that forwards to the original abstract method. 

- If the new abstract method is a variant of a previousl6 existing abstract method, just make it return some reasonable default value

mrdziuban:
Thanks for the tips, all the changes are binary compatible now 🎉 

I'll hopefully be able to add some tests over the next couple of days

mrdziuban:
Added a couple tests, I think this is ready for review now

upickle/test/src/upickle/ClassDefs.scala
```diff
@@ -55,7 +55,7 @@ object ADTs {
 object Hierarchy {
   sealed trait A
   object A{
-    implicit def rw: upickle.default.ReadWriter[A] = RW.merge(B.rw, C.rw)
+    implicit def rw: upickle.default.ReadWriter[A] = RW.merge(core.Annotator.defaultTagKey, B.rw, C.rw)
```
lihaoyi, line 58: For backwards compatibility, could we call the new method `RW.mergeKeyed` or something? 

I think it's reasonable to leave the existing functionality with the default in place rather than forcing people to explicitly pass in a `defaultTagKey` every time they call `merge`, and because of the use of varargs we can't give `defaultTagKey` a default value. So having a separate method `merge` vs `mergeKeyed` seems like the way to go

lihaoyi:
Left one code comment. One more request, could you update the documentation site? You can use `sbt upickleReadme/run` to re-generate it in the `upickleReadme/target/scalatex` folder

lihaoyi:
Also, could you update the PR description? A few lines summarizing what code changes are necessary, any alternatives you considered, and any other concerns you thought about, should be enough

upickle/test/src/upickle/ClassDefs.scala
mrdziuban: Definitely, updated

mrdziuban:
> could you update the documentation site to include the new functionality?

👍 done

> could you update the PR description?

For sure, updated to add these details

lihaoyi:
Looks good to me. I kicked off CI and will merge it when green. Send me your bank details via email to haoyi.sg@gmail.com and i will close out the bounty

mrdziuban:
Whoops sorry about the test failures, fixing in a minute

lihaoyi:
Thanks!

PR #590, opened 2024-06-18
Change bitwise OR to boolean OR to prevent Chrome ARM issue

matthewdunbar:


ekrich:
Not sure if this is related but ... https://stackoverflow.com/questions/27583255/javascript-why-does-bitwise-or-of-booleans-return-a-number-instead-of-a-boolean

PR #613, opened 2024-07-21
Add Float32 and Float64 to Msg conversions

ramon54321:
Small change to include `Float32` and `Float64` conversions in the `Msg` class. Currently there is no way to extract floats from `Msg` without a cast at the call site.


ramon54321:
Hello, how does this work, can this be merged?

PR #631, opened 2024-09-19
Fix Readme.scalatex

nox213:
There is a compile error in the documentation generation code.
Because of that I believe documentation was not updated after the last release.


lihaoyi:
Fixed in master, thanks for the report!

PR #634, opened 2024-09-26
Fix macro crash when handling sealed case classes

### LINKED ISSUE #628: Sealed case class causes compiler crash
```scala
sealed case class ImportType(name: String)
implicit val importTypeRW: upickle.default.ReadWriter[ImportData.ImportType] =
    upickle.default.macroRW
```
```
[error] /Users/lihaoyi/Github/Ammonite/amm/runtime/src/main/scala/ammonite/runtime/Storage.scala:114:21: exception during macro expansion:
[error] java.lang.UnsupportedOperationException: empty.max
[error] 	at scala.collection.IterableOnceOps.max(IterableOnce.scala:1127)
[error] 	at scala.collection.IterableOnceOps.max$(IterableOnce.scala:1124)
[error] 	at scala.collection.AbstractIterable.max(Iterable.scala:935)
[error] 	at upickle.implicits.internal.Macros$DeriveDefaults.annotate(Macros.scala:231)
[error] 	at upickle.implicits.internal.Macros$DeriveDefaults.annotate$(Macros.scala:209)
[error] 	at upickle.implicits.internal.Macros$Reading.annotate(Macros.scala:278)
[error] 	at upickle.implicits.internal.Macros$DeriveDefaults.deriveClass(Macros.scala:203)
[error] 	at upickle.implicits.internal.Macros$DeriveDefaults.deriveClass$(Macros.scala:160)
[error] 	at upickle.implicits.internal.Macros$Reading.deriveClass(Macros.scala:278)
[error] 	at upickle.implicits.internal.Macros$DeriveDefaults.derive(Macros.scala:115)
[error] 	at upickle.implicits.internal.Macros$DeriveDefaults.derive$(Macros.scala:109)
[error] 	at upickle.implicits.internal.Macros$Reading.derive(Macros.scala:278)
[error] 	at upickle.implicits.internal.Macros$.macroRImpl(Macros.scala:473)
[error]     upickle.default.macroRW
[error]                     ^
```

Fails on most Scala versions. It seem `sealed` is confusing the macro into thinking it's a `sealed trait`, and then when it finds no `case class` or `case object` children it crashes

Workaround is to remove the `sealed`, but really we should just not crash

nox213:
Fixes #628

 Sealed classes can be instantiated directly. Therefore, when generating the typeTag, I included the sealed class itself. This ensures that the macro correctly handles sealed classes without subclasses. To check whether it’s a sealed class and not a trait or abstract class, I used the following condition:
 ```
 sealedParents.find(_ == tpe.typeSymbol)
```
Since `trait` and `abstract class` cannot be instantiated, I believe this approach works. However, I could also explicitly check that the symbol is not a trait or an abstract class, if needed.

What do you think?

### DIFF
--- a/upickle/implicits/src-2/upickle/implicits/internal/Macros.scala
+++ b/upickle/implicits/src-2/upickle/implicits/internal/Macros.scala
@@ -219,9 +219,9 @@ object Macros {
           fail(tpe, _),
         )
 
-        val myself: Option[Symbol] = sealedParents.find(_ == tpe.typeSymbol)
+        val sealedClassSymbol: Option[Symbol] = sealedParents.find(_ == tpe.typeSymbol)
         val segments =
-          myself.toList.map(_.fullName.split('.')) ++
+          sealedClassSymbol.toList.map(_.fullName.split('.')) ++
             sealedParents
             .flatMap(_.asClass.knownDirectSubclasses)
             .map(_.fullName.split('.'))
--- a/upickle/implicits/src-2/upickle/implicits/internal/Macros.scala
+++ b/upickle/implicits/src-2/upickle/implicits/internal/Macros.scala
@@ -219,11 +219,13 @@ object Macros {
           fail(tpe, _),
         )
 
-
+        val myself: Option[Symbol] = sealedParents.find(_ == tpe.typeSymbol)
         val segments =
-          sealedParents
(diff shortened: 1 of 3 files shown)


lihaoyi:
@nox213 looks great, thanks!

PR #641, opened 2024-11-14
WIP: Named tuples

Quafadas:
Compiling fails with this;

```
[1/1] show
[1-148/162] upickle_named_tuples.jvm[3.6.1].compile
[1-148] [info] compiling 1 Scala source to /Users/simon/Code/upickle/out/upickle_named_tuples/jvm/3.6.1/compile.dest/classes ...
[1-148] [error] -- [E008] Not Found Error: /Users/simon/Code/upickle/upickle_named_tuples/src/NamedTupleReadWriter.scala:4:15 
[1-148] [error] 4 |import upickle.default.*
[1-148] [error]   |       ^^^^^^^^^^^^^^^
[1-148] [error]   |       value default is not a member of upickle
[1-148] [error] -- [E046] Cyclic Error: /Users/simon/Code/upickle/upickle_named_tuples/src/NamedTupleReadWriter.scala:8:45 
[1-148] [error] 8 |  inline given [N <: Tuple, V <: Tuple, T <: NamedTuple[N, V]]: ReadWriter[T] =
[1-148] [error]   |                                             ^
[1-148] [error]   |                               Cyclic reference involving val <import>
[1-148] [error]   |
[1-148] [error]   |                                Run with -explain-cyclic for more details.
[1-148] [error]   |
[1-148] [error]   | longer explanation available when compiling with `-explain`
[1-148] [error] -- [E008] Not Found Error: /Users/simon/Code/upickle/upickle_named_tuples/src/NamedTupleReadWriter.scala:3:13 
[1-148] [error] 3 |import scala.NamedTuple.NamedTuple
[1-148] [error]   |       ^^^^^^^^^^^^^^^^
[1-148] [error]   |       value NamedTuple is not a member of scala
[1-148] [error] three errors found

```

Which makes me think I've stuffed up the build horridly, as it apparently doesn't resolve either pickle or named tuples. 

I can't see how tho.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ out/
 *.iml
 .idea
 .DS_Store
+.bloop
+.metals
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+  "files.watcherExclude": {
+    "**/target": true
+  }
+}
--- a/build.mill
+++ b/build.mill
@@ -391,26 +391,23 @@ trait BenchModule extends CommonPlatformModule{
 object upickle_named_tuples extends Module {
   // todo: replace with scala3 version when upickle moves to next LTS
   val scalaVersionWithNamedTuples = Seq("3.6.1")
-  trait CommonCoreModule extends CommonPublishModule {
-    def ivyDeps = Agg(ivy"com.lihaoyi::geny::1.1.1")
-  }
 
   object js extends Cross[CoreJsModule](scalaVersionWithNamedTuples)
-  trait CoreJsModule extends CommonJsModule with CommonCoreModule {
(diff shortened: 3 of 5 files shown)


Quafadas:
I believethat modules can't easily depend on different scala 3 versions, which makes this PR painful to do in a reasonable way. 

Conclusion, wait until scala has an LTS > 3.6.x

lolgab:
@Quafadas I think the problem was the @experimental annotation on `NamedTuple`s in Scala 3.6.
I opened a PR to your branch here: https://github.com/Quafadas/upickle/pull/1

lolgab:
Reporting my message here:

> I think maybe it needs to be published in its own separate module or something whilst we wait for another LTS

That's necessary. Maybe it can be an experimental module and can live in a separate package, and then when it becomes LTS it can be moved under the normal `import upickle.default.*`.
Something like `upickle-next`.

PR #661, opened 2025-04-19
chore: override foreach in LinkedHashMap

He-Pin:
Motivation:
override it can improve some performance as it is been used in sjsonnet


lihaoyi:
Can you show benchmarks demonstrating the performance increase?

He-Pin:
Not prepared one,but `forEach()` can be implemented to be faster than for-each loop, because the iterable knows the best way to iterate its elements, as opposed to the standard iterator way. So the difference is loop internally or loop externally.

lihaoyi:
Without benchmarks we'll probably not merge this. Otherwise we have no idea how this will affect performance and may well make things slower

PR #678, opened 2025-09-02
Make getElemSafe throw EOFException on misuse

### LINKED ISSUE #672: BufferingByteParser.getByteSafe is not safe (in upickle.core)
The key is this method below, which does not check the return value of `requestUntil`, and as a result, upack behaves strangely in the context of EOFs.

```scala
def getByteSafe(i: Int): Byte = {
  requestUntil(i)
  buffer(i - firstIdx)
}
```

I ran into this issue when trying to parse a file containing contiguous MsgPack records.
For newline delimited JSON you can split the file into lines and parse those, but MsgPack doesn't have lines.
So, I needed to actually make multiple parse calls using the same input stream, and all my attempts were broken in strange ways.

I initially tried `upack.read(inputStream)`, which reads the first record from the file, and then infinitely reads int 0.
On examination:
- It makes sense that read 2 failed, since there's a hidden buffer construction which emptied the input stream. That's fine, makes sense.
- The infinite 0 part was weird, though. Surely it would throw an exception on an empty stream?
- Turns out msgpack has "negative fixint", which maps the value `-1` to the int `0`. Given that InputStreams return -1 on EOF, maybe that's it?

In any case, I moved to a code example that in my opinion should work, but doesn't because of the lack of EOF checking.
```scala
val reader = new upack.InputStreamMsgPackReader(inputStream)
while true
do reader.parse(upickle.default.reader[TraceRecord])
```
This moves the buffer construction outside of the repeated parsing, and does correctly parse all the records in the file.
What doesn't work is the EOF check, which in this case throws an `ArrayIndexOutOfBoundsException` when the input is exhausted.

If I override that one specific method, I can get a more useful behavior:
```scala
val reader = new upack.InputStreamMsgPackReader(inputStream)/l
  override def getByteSafe(i: Int): Byte =
    if requestUntil(i)
    then throw java.io.EOFException("EOF")
    else super.getByteSafe(i)
while true
do reader.parse(upickle.default.reader[TraceRecord])
```

In this case, the program ends with `EOFException`, which, though I just chose it arbitrarily, seems like a more useful error condition than having the parser misread -1, or throwing an array bounds check error.

There might be some other design reason behind the missing check, or perhaps it's just an ignored return value.
All else being equal, I'll try to eventually PR a change with my proposed fix and additional tests.

fhackett:
Fixes #672.

Without the change to BufferingElemParser, the test will misread int 0 when it goes beyond the end of the input array.
Adding a bounds check lets us see an exception indicating the true situation, rather than reading corrupt data.

This change makes parsing concatenated sequences of MsgPack values possible without a tricky workaround.
It should not affect usage where only one MsgPack value is expected.

The one possible change this may cause for users is when parsing a truncated MsgPack value, in which case upack should more reliably throw `java.io.EOFException`, rather than some implementation related error, or in some theoretical scenario, parse a corrupted MsgPack value (I can add a test for this if that's of interest).

I'm also hoping CI will run on this and expose any issues with Scala Native or Scala JS - I had issues testing those locally.

### DIFF
--- a/upack/src/upack/MsgPackReader.scala
+++ b/upack/src/upack/MsgPackReader.scala
@@ -8,9 +8,11 @@ class MsgPackReader(input0: Array[Byte]) extends BaseMsgPackReader {
   val srcLength = input0.length
   protected[this] final def close() = {}
 
-  // Make sure we never call this method, since it will mutate the original array,
-  // and it should not be necessary to call it if our implementation is correct.
-  override def growBuffer(until: Int): Unit = ???
+  // If we call this method, assuming our implementation is correct,
+  // we are processing a truncated input. Throw EOFException to signal this
+  // to our caller.
+  override def growBuffer(until: Int): Unit =
+    throw new java.io.EOFException(s"index $until is outside the range of input byte array (length $srcLength)")
 
   def readDataIntoBuffer(buffer: Array[Byte], bufferOffset: Int) = {
     if(buffer == null) (input0, false, srcLength)
--- a/upack/test/src/upack/UnitTests.scala
+++ b/upack/test/src/upack/UnitTests.scala
@@ -105,14 +105,18 @@ object UnitTests extends TestSuite{
       val bytes = upack.write(msg)
       val truncatedBytes = bytes.take(2)
 
-      // Uses upack.MsgPackReader (which has its own growBuffer override that throws)
-      intercept[java.io.EOFException]{
-        upack.read(truncatedBytes)
+      // upack.MsgPackReader has its own growBuffer override that throws
+      test("upack.MsgPackReader"){
+        intercept[java.io.EOFException]{
+          upack.read(truncatedBytes) ==> msg
+        }
       }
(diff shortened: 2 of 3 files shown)


fhackett:
Whoops, accidentally used Scala 3 only syntax in my change. Should be fixed now.

fhackett:
@lihaoyi I added the extra test. I confirmed that, when using the InputStream reader, the tests I just pushed would reparse a corrupted `Arr(1, 0, 0, 0)` from a truncated serialization of `Arr(1, 2, 3, 4)` (and therefore fail).

I also stumbled on the special case for byte arrays, which hit `???` instead due to its `growBuffer` override. I changed it to throw `EOFException` instead, which is more in line with what I'm proposing. I'm guessing that `???` is there to catch an implementation error, whereas, given a correct implementation, the only other way to hit `growBuffer` is if the input array is truncated (or I guess if the MsgPack input is malformed in a way that suggests the input is longer than it really is), in which case a user might appreciate flagging it as `EOFException`.

Let me know what you think, and/or if you have any other things you'd like me to try.

PR #679, opened 2025-09-05
feat: string-based-union support

ThijsBroersen:


lihaoyi:
I think the approach looks fine as a first draft, but my concern is how it scales to broader usages of union types? Apart from unions of literal `String` types, can we also support unions of different primitives (e.g. `String | Int | Boolean`), or unions of different distinct `case class`es or `object`s?

I mainly to make sure there's some coherent approach that we can document and users can understand, and that whatever we implement now doesn't get in the way of implementing other union types in future

PR #681, opened 2025-09-10
Cross compile ujson.circe to Scala JS

majk-p:


build.mill
```diff
@@ -227,6 +227,13 @@ object ujson extends Module{
       def mvnDeps = Seq(mvn"io.circe::circe-parser:$circeVersion")
     }
 
+    object js extends Cross[CirceJsModule](scalaVersions)
+    trait CirceJsModule extends CommonJsModule{
+      def moduleDeps = Seq(ujson.js())
+      val circeVersion = "0.14.14"
+      def mvnDeps = Seq(mvn"io.circe::circe-parser:$circeVersion")
```
lolgab, line 234: Can you extract these in a trait as we do in `CommonCoreModule` and then extend it in jvm, js and native?

lolgab (changes requested):

build.mill
majk-p: Sure thing, done

PR #723, opened 2026-07-11
Reject unpaired JSON surrogate escapes

### LINKED ISSUE #722: ujson accepts unpaired JSON surrogate escapes as raw UTF-16 surrogates
## Summary

ujson 4.4.3 accepts JSON string escapes containing unpaired UTF-16 surrogate halves and exposes them as raw `String` contents. For example, parsing `"\\uD800"` yields a one-code-unit string whose first char is `0xD800`.

That makes downstream libraries observe invalid Unicode scalar values instead of getting a JSON parse failure or a replacement character. This is especially awkward for consumers whose string semantics are defined in terms of Unicode code points/scalar values.

## Reproducer

```scala
val value = ujson.read("\"\\uD800\"")
val s = value.str
println(s.length)                 // 1
println(s.charAt(0).toInt)         // 55296 / 0xD800
```

The same applies to lone low surrogates and to object keys:

```scala
ujson.read("\"\\uDC00\"").str.charAt(0).toInt
ujson.read("{\"\\uD800\": 1}").obj.keys.head.charAt(0).toInt
```

Valid surrogate pairs should continue to decode normally:

```scala
ujson.read("\"\\uD83D\\uDE00\"").str.codePointAt(0) // 128512
```

## Expected Behavior

Ideally ujson should reject unpaired surrogate escapes while decoding JSON strings, at the point where it already has high/low surrogate context. That avoids downstream validation passes over every decoded string.

A replacement-character policy would also be safer than preserving raw surrogate halves, but rejection is easier for strict JSON consumers to reason about and matches parsers such as nlohmann-json and serde_json.

## Downstream Impact

sjsonnet currently depends on `com.lihaoyi::ujson::4.4.3` for `std.parseJson` and `.json` imports. Jsonnet strings and `std.codepoint` operate in Unicode-codepoint terms; preserving raw surrogate halves lets `std.parseJson("\"\\uD800\"")` expose `55296`, while sjsonnet source string literals already reject the equivalent `"\\uD800"` escape.

A ujson-level fix would let sjsonnet remove its downstream workaround and pick up the behavior by upgrading ujson.

## References

- RFC 8259 section 8.2 notes that unpaired UTF-16 surrogates lead to unpredictable behavior: https://www.rfc-editor.org/rfc/rfc8259#section-8.2
- Jsonnet `std.parseJson` documentation: https://jsonnet.org/ref/stdlib.html#std-parseJson

He-Pin:
## Summary

- Reject lone high and low UTF-16 surrogate escapes while parsing JSON strings
- Require a high surrogate escape to be followed immediately by a low surrogate escape
- Preserve valid surrogate pairs across the generated Char and Byte parser backends
- Add regression coverage for string values, object keys, parser/render/validate entry points, input forms, and 2-byte stream buffers

## Motivation

ujson currently exposes unpaired `\uD800`-`\uDFFF` escapes as raw UTF-16 surrogate halves. This leaks invalid Unicode scalar values to downstream consumers. In particular, sjsonnet uses ujson for `std.parseJson` and `.json` imports, while Jsonnet string/codepoint semantics expect Unicode code points.

The validation belongs in ujson's shared string parser, where both halves are available and all values, keys, input forms, and renderers pass through the same path. This avoids downstream post-validation scans.

## Behavior

- `"\uD83D\uDE00"` continues to parse as a valid surrogate pair
- `"\uD800"`, `"\uDC00"`, and mismatched pairs now fail with `ParsingFailedException`
- The same validation applies to object keys

Fixes #722

## Testing

- `./mill 'ujson.jvm[2.13.16].test.testForked'` — 248/248 successful, including the 5 GB streaming test
- `./mill 'ujson.jvm[3.3.7].test.testOnly' ujson.SyntaxTests` — 82/82 successful
- `./mill 'ujson.js[2.13.16].test.testOnly' ujson.SyntaxTests` — 82/82 successful
- Qoder CLI blocker review — no must-fix findings

### DIFF
--- a/ujson/templates/ElemParser.scala
+++ b/ujson/templates/ElemParser.scala
@@ -262,8 +262,8 @@ abstract class ElemParser[J] extends upickle.core.BufferingElemParser{
   /**
    * Generate a Char from the hex digits of "\u1234" (i.e. "1234").
    *
-   * NOTE: This is only capable of generating characters from the basic plane.
-   * This is why it can only return Char instead of Int.
+   * NOTE: Returns a single UTF-16 code unit, which may be a surrogate
+   * (D800-DFFF). Callers are responsible for validating surrogate pairs.
    */
   protected[this] final def descape(i: Int): Char = {
     import upickle.core.RenderUtils.hex
@@ -581,7 +581,10 @@ abstract class ElemParser[J] extends upickle.core.BufferingElemParser{
 
           case 'u' => {
             val escaped = descape(i)
-            if (Character.isHighSurrogate(escaped)) {
+            if (!Character.isSurrogate(escaped)) {
+              outputBuilder.appendC(escaped)
+              i += 6
+            } else if (Character.isHighSurrogate(escaped)) {
               if (getElemSafe(i + 6) != '\\' || getElemSafe(i + 7) != 'u') {
                 die(i + 6, "expected \\u escape after high surrogate")
               }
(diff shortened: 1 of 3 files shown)


He-Pin:
@lolgab wdyt

